PR #12, opened 2022-06-20
CI

SeungheonOh:


SeungheonOh:
@peter-mlabs We need to set CACHIX_KEY. I don't have it. Once we have that we'll have a simple CI that checks if it runs.

PR #14, opened 2022-06-20
`PFixedDecimal`

### LINKED ISSUE #13: Fixed Width Decimal
Fixed width decimal with unit of 1,000,000. 

It would be a simple wrapper around regular integer that provides numeric operations specifically for fixed width decimal. Main goal is to handle the Lovelace values, so a converting interfaces to Value is also required.

SeungheonOh:
close #13

### DIFF
--- a/src/Plutarch/Extra/FixedDecimal.hs
+++ b/src/Plutarch/Extra/FixedDecimal.hs
@@ -1,0 +1,80 @@
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE ViewPatterns #-}
+
+module Plutarch.Extra.FixedDecimal (
+    PFixedDecimal (..),
+    DivideSemigroup (..),
+    DivideMonoid (..),
+    decimalToAdaValue,
+) where
+
+import Plutarch.Api.V1
+import Plutarch.Api.V1.Value
+import Plutarch.Bool
+import Plutarch.Integer
+import Plutarch.Prelude
+import Plutarch.Unsafe
+
+decimalUnit :: Term s PInteger
+decimalUnit = 1000000
+
+{- | Fixed width decimal. Decimal point is at 1,000,000.
+ This would be used for representing Ada value with some Lovelace changes.
+
+ @since 1.0.0
+-}
+newtype PFixedDecimal (s :: S)
+    = PFixedDecimal (Term s PInteger)
+
+-- | @since 1.0.0
+deriving via
+    (DerivePNewtype PFixedDecimal PInteger)
+    instance
+        (PlutusType PFixedDecimal)
+
+-- | @since 1.0.0
+deriving via
+    (DerivePNewtype PFixedDecimal PInteger)
+    instance
+        PEq PFixedDecimal
+
+-- | @since 1.0.0
+deriving via
+    (DerivePNewtype PFixedDecimal PInteger)
+    instance
+        POrd PFixedDecimal
+
+-- | @since 1.0.0
+instance Num (Term s PFixedDecimal) where
+    (+) = (+)
+    (-) = (-)
+    (punsafeCoerce -> x) * (punsafeCoerce -> y) =
+        punsafeCoerce $ pdiv # (x * y) # decimalUnit
+    abs = abs
+    signum = signum
+    fromInteger = punsafeCoerce . (* decimalUnit) . pconstant
+
+-- TODO: This should be moved to either to plutarch-numeric or other module
+class DivideSemigroup a where
+    divide :: a -> a -> a
+
+class (DivideSemigroup a) => DivideMonoid a where
+    one :: a
+
+-- | @since 1.0.0
+instance DivideSemigroup (Term s PFixedDecimal) where
+    divide (punsafeCoerce -> x) (punsafeCoerce -> y) =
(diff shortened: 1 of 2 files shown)


src/Plutarch/Extra/FixedDecimal.hs
```diff
@@ -0,0 +1,80 @@
...
+    abs = abs
+    signum = signum
+    fromInteger = punsafeCoerce . (* decimalUnit) . pconstant
+
+-- TODO: This should be moved to either to plutarch-numeric or other module
+class DivideSemigroup a where
+    divide :: a -> a -> a
+
+class (DivideSemigroup a) => DivideMonoid a where
+    one :: a
```
t1lde, line 62: I wonder whether these numeric classes are a good idea, rather than something more like purescript's `DivisionRing`. :thinking:

src/Plutarch/Extra/FixedDecimal.hs
```diff
@@ -0,0 +1,80 @@
...
+        POrd PFixedDecimal
+
+-- | @since 1.0.0
+instance Num (Term s PFixedDecimal) where
+    (+) = (+)
+    (-) = (-)
+    (punsafeCoerce -> x) * (punsafeCoerce -> y) =
+        punsafeCoerce $ pdiv # (x * y) # decimalUnit
+    abs = abs
+    signum = signum
```
t1lde, line 54: ?

t1lde: won't these bottom?

src/Plutarch/Extra/FixedDecimal.hs
```diff
@@ -0,0 +1,80 @@
...
+    (DerivePNewtype PFixedDecimal PInteger)
+    instance
+        POrd PFixedDecimal
+
+-- | @since 1.0.0
+instance Num (Term s PFixedDecimal) where
+    (+) = (+)
+    (-) = (-)
+    (punsafeCoerce -> x) * (punsafeCoerce -> y) =
+        punsafeCoerce $ pdiv # (x * y) # decimalUnit
```
t1lde, line 52: should prefer more specific coercions (pto, pcon in this case I think) where they are available - it's slightly safer

src/Plutarch/Extra/FixedDecimal.hs
```diff
@@ -0,0 +1,80 @@
...
+
+import Plutarch.Api.V1
+import Plutarch.Api.V1.Value
+import Plutarch.Bool
+import Plutarch.Integer
+import Plutarch.Prelude
+import Plutarch.Unsafe
+
+decimalUnit :: Term s PInteger
+decimalUnit = 1000000
```
t1lde, line 19: I don't like this being hardcoded so much - since we have plutarch we can think about doing better.
I have some ideas:

- we could have the decimal unit be a type-level Nat, which `PFixedDecimal` is tagged with. 
- then (in combination with `plutarch-safe-money`!? we could define a `Unit` typeclass, which defines the 'base' of various units, which we can then use to define typed unit conversions!

src/Plutarch/Extra/FixedDecimal.hs
t1lde: maybe it's a special syntax for using a default implementation, idk.

src/Plutarch/Extra/FixedDecimal.hs
t1lde: let me whip up an example, as I think this is also adjacent to @tbidne 's time unit handling

SeungheonOh: This sounds really nice, I'll put typelevel nat as unit decimal

src/Plutarch/Extra/FixedDecimal.hs
SeungheonOh: I should as @kozross about this. I can later change this to follow what he implements on `plutarch-numeric`

t1lde (approved):
very nice :)

SeungheonOh:
I want to try something. I'll merge it after that :+1:

src/Plutarch/Extra/FixedDecimal.hs
```diff
@@ -0,0 +1,95 @@
...
+
+-- | @since 1.0.0
+decimalToAdaValue ::
+    forall (s :: S) (keys :: KeyGuarantees) (amounts :: AmountGuarantees) (unit :: Nat).
+    KnownNat unit =>
+    Term s (PFixedDecimal unit :--> PValue keys amounts)
+decimalToAdaValue =
+    phoistAcyclic $
+        plam $ \(pto -> dec) ->
+            let adaValue = (pdiv # dec # (pconstant (natVal (Proxy @unit)))) * pconstant 1000000
```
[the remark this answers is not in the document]
SeungheonOh: `PFixedDecimal` is in Ada-unit, but with capability to have decimal amounts. 
So if `PFixedDecimal' have 12.345 and run `decimalToAdaValue` it will give 12,345,000 Lovelace.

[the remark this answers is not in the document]
t1lde: this makes sense, actually :+1:

t1lde (approved):
lgtm :+1:

PR #20, opened 2022-06-23
Additional utilities for `PFixedDecimal` to build `PExchangeRate`

SeungheonOh:


src/Plutarch/Extra/FixedDecimal.hs
```diff
@@ -93,3 +141,47 @@ decimalToAdaValue =
...
+{- | Convert @PInteger@ to @PFixedDecimal@ in Haskell level.
+
+ @since 1.0.0
+-}
+fromPInteger' ::
+    forall (unit :: Nat) (s :: S).
+    KnownNat unit =>
+    Term s PInteger ->
+    Term s (PFixedDecimal unit)
+fromPInteger' z = pcon . PFixedDecimal $ pconstant (natVal (Proxy @unit)) * z
```
kozross, line 165: Is there a reason we need both of these? I feel we can just have the Plutarch-level function only, and then use `#` in cases where we need a `Term`-based function.

src/Plutarch/Extra/FixedDecimal.hs
```diff
@@ -93,3 +141,47 @@ decimalToAdaValue =
...
+{- | Identical to @toPInteger@ but Haskell level.
+
+ @since 1.0.0
+-}
+toPInteger' ::
+    forall (unit :: Nat) (s :: S).
+    KnownNat unit =>
+    Term s (PFixedDecimal unit) ->
+    Term s PInteger
+toPInteger' d = pdiv # pto d # (pconstant (natVal (Proxy @unit)))
```
kozross, line 187: Same as above.

kozross (changes requested):

src/Plutarch/Extra/FixedDecimal.hs
SeungheonOh: Looking back, we really don't them these.

kozross (approved):

PR #22, opened 2022-06-24
add address utils, add entr to flake (for 'make watch')

peter-mlabs:
more functions for making addresses

### DIFF
--- a/src/Plutarch/Api/V1/ScriptContext.hs
+++ b/src/Plutarch/Api/V1/ScriptContext.hs
@@ -234,6 +234,7 @@ paddressFromValidatorHash = plam $ \valHash stakingCred ->
         pdcons # pdata (pcon $ PScriptCredential (pdcons # pdata valHash # pdnil))
             #$ pdcons # pdata stakingCred
             #$ pdnil
+
 {- | Constuct an address (with a staking credential) from a @PPubKeyHash@
 and maybe a @PStakingCredential
 
--- a/src/Plutarch/Api/V1/ScriptContext.hs
+++ b/src/Plutarch/Api/V1/ScriptContext.hs
@@ -3,6 +3,10 @@
 {-# LANGUAGE ViewPatterns #-}
 
 module Plutarch.Api.V1.ScriptContext (
+    paddressFromValidatorHash,
+    paddressFromValidatorHash',
+    paddressFromPubKeyHash,
+    paddressFromPubKeyHash',
     pownTxOutRef,
     pownTxInfo,
     pownValue,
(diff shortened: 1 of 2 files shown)


src/Plutarch/Api/V1/ScriptContext.hs
```diff
@@ -216,3 +222,55 @@ pvalidatorHashFromAddress = phoistAcyclic $
         pmatch (pfromData $ pfield @"credential" # addr) $ \case
             PScriptCredential ((pfield @"_0" #) -> vh) -> pcon $ PJust vh
             _ -> pcon $ PNothing
+
+{- | Construct an address (without a staking credential) from a @PValidatorHash@
+
+    @since 1.1.0
+-}
+paddressFromValidatorHash ::
```
SeungheonOh, line 230: Can we combine `paddressFromValidatorHash'` and `paddressFromValidatorHash` with something like `Term s (PValidatorHash :--> PMaybe PStakingCredential :--> PAddress)`?

src/Plutarch/Api/V1/ScriptContext.hs
```diff
@@ -216,3 +222,55 @@ pvalidatorHashFromAddress = phoistAcyclic $
...
+    pcon . PAddress $
+        pdcons # pdata (pcon $ PScriptCredential (pdcons # pdata valHash # pdnil))
+            #$ pdcons # pdata (pcon $ PDJust (pdcons # pdata stakingCred # pdnil))
+            #$ pdnil
+
+{- | Constuct an address (without a staking credential) from a @PPubKeyHash@
+
+    @since 1.1.0
+-}
+paddressFromPubKeyHash ::
```
SeungheonOh, line 256: Same thing.

SeungheonOh (changes requested):

src/Plutarch/Api/V1/ScriptContext.hs
peter-mlabs: I'm open to this, but its the difference between potentially doing a `pcon . PDNothing $ (pdnil)` or not, which seems annoying. What do you think?

SeungheonOh: It will be `pcon PNothing`. It would still look a little be ugly, but I think having one function is much cleaner.

peter-mlabs: Nah, it's definitely `pcon . PDNothing $ pdnil`

But either way, I think its nbd. Fixed.

peter-mlabs: that works, @chfanghr! I always forget that that's allowed...

Issue #23, opened 2022-06-24
Add construction/deconstruction of `PTagged` values to haddocks


peter-mlabs:
We could use some usage guidelines/examples for (de)constructing ptagged values. 

- Construction seems like it should be done with `pconstant`; does `pcon` work as well? Also seems like we typically need type annotations or application; I don't have a strong preference, but we should make a decision on what we want to consider best practice.
- Deconstruction can happen via `pmatch`, `pto`, or `pextract`. No preference here either, but we should pick one and add it.

We should also try to verify that the performance of each of these are the same.

PR #24, opened 2022-06-27
Little guidance for `PTagged`

SeungheonOh:


SeungheonOh:
close #23

src/Plutarch/Extra/Tagged.hs
```diff
@@ -48,7 +48,56 @@ import Plutarch.TryFrom (PTryFrom (PTryFromExcess, ptryFrom'))
 import Plutarch.Unsafe (punsafeCoerce)
 import qualified PlutusTx
 
-{- | Plutarch-level 'Tagged'.
+{- | Plutarch-level 'Tagged'. @PTagged@ allows one to "tag" a given type in the typelevel.
```
kozross, line 51: Usually, it's 'at the type level'.

src/Plutarch/Extra/Tagged.hs
```diff
@@ -48,7 +48,56 @@ import Plutarch.TryFrom (PTryFrom (PTryFromExcess, ptryFrom'))
...
+ Typically, @tag@ will hold `Symbol` or custom kind.  Since @PTagged@s only act as a wrapper, it does not hold any significance in the resulting script, and there are no affect on performance/script size.
```
kozross, line 52: "will be a 'Symbol', or some other custom kind"

"Since @PTagged@s only act as wrappers, their \'tags\' don't hold any significance in the resulting script, and have no effect on performance or script size."

src/Plutarch/Extra/Tagged.hs
```diff
@@ -48,7 +48,56 @@ import Plutarch.TryFrom (PTryFrom (PTryFromExcess, ptryFrom'))
...
+
+=== Tagged Types:
+@
+-- One can make new kind with
```
kozross, line 56: "We can create a new kind by defining an empty type. This is then \'promoted\' by GHC to a kind."

src/Plutarch/Extra/Tagged.hs
```diff
@@ -48,7 +48,56 @@ import Plutarch.TryFrom (PTryFrom (PTryFromExcess, ptryFrom'))
...
+data Ada
+data GT
+
+PTagged Ada PInteger
+PTagged GT PValue
+
+-- Or one can using Symbol, typelevel string
```
kozross, line 63: "Or we can use 'Symbol' (a type-level string)"

src/Plutarch/Extra/Tagged.hs
```diff
@@ -166,7 +215,11 @@ instance (PConstantDecl a) => PConstantDecl (Tagged t a) where
     pconstantToRepr (Tagged x) = pconstantToRepr x
     pconstantFromRepr x = Tagged <$> pconstantFromRepr x
 
--- | @since 1.0.0
+{- | Retags given @PTagged@. The underlaying type of input and output must be equal.
```
kozross, line 218: The second sentence is redundant; the type already tells us this.

kozross (changes requested):

kozross (approved):

PR #31, opened 2022-07-12
Traversable improvements

kozross:
The specifics are detailed in the changelog. This improves performance (potentially significantly) when `PTraversable` and `PSemiTraversable` are used for fold-like behaviours only. We also introduce some additional flexibility for 'placeholder' types via `PBoring`.


peter-mlabs (approved):

peter-mlabs (approved):
Looks good now, once CI passes.

PR #32, opened 2022-07-12
add hydraJobs to flake.nix

peter-mlabs:


SeungheonOh:
Hydra is working! I don't think this PR is needed.

PR #41, opened 2022-07-24
PBind, PStar, instances

kozross:
This adds a parallel notion to `Bind` from `semigroupoids` at the Plutarch level. This, together with `PApplicative`, gives us `PMonad` in essence. I've defined instances everywhere sensible: in particular, I didn't bother with instances for `PTagged` and `PSum`, even though technically there's nothing prohibiting this, as I don't think they'd be used much in a monadic context anyway.

I've also added `PStar` as a direct counterpart to `Kleisli` (and, indeed, `Star` from `profunctors`), as it has nice instances, especially of `PSemigroupoid` and `PCategory`. I didn't define its `PBind` instance (although it can have one) mostly because I don't expect us to need it much, and the type tetris required to make it play nice with the representation polymorphism we have for Plutarch 'containery' things is tedious.


peter-mlabs (commented):
What is `PStar` actually useful for? 

I think we should add a module docstring for `PBind`, since the word "monad" doesn't appear anywhere in the module; it's not very greppable.

kozross:
@peter-mlabs `PStar` gives us the ability to compose `a :--> f b` in the same way we can compose `a :--> b`; `a :--> b` is an arrow in `Plut`, whereas `a :--> f b` is an arrow in the Kleisli category defined by `f`. Therefore, if `f` is a `PBind`, `PStar f` is a semigroupoid (and indeed, a `PSemigroupoid`), while if `f` is _also_ a `PApplicative`, `PStar f` is a category (and indeed, a `PCategory`).

In short, it means we don't need (the equivalents of) both `.` and `<=<`, since `#>>>` does _both_ duties.

src/Plutarch/Extra/Bind.hs
```diff
@@ -0,0 +1,140 @@
...
+            xs' <- pmatchC xs
+            pure $ case xs' of
+                PDNothing t -> pcon . PDNothing $ t
+                PDJust t -> f # pfromData (pfield @"_0" # t)
+
+-- | @since 1.2.1
+instance PBind PList where
+    pbind = phoistAcyclic $
+        pfix #$ plam $ \self xs f -> unTermCont $ do
+            xs' <- pmatchC (puncons # xs)
```
cjay, line 82: The temporary `PMaybe` here is not necessary, it's probably more efficient to directly match on `PSCons` and `PSNil`. I suppose `puncons` sholud only be used when being polymorphic in the list type.

src/Plutarch/Extra/Bind.hs
```diff
@@ -0,0 +1,140 @@
...
+ @'pliftA2' '#' 'papply' '#' f '#' x@
+
+ @since 1.2.1
+-}
+class (PApply f) => PBind (f :: (S -> Type) -> S -> Type) where
+    -- | '>>=', but as a Plutarch function.
+    pbind ::
+        forall (a :: S -> Type) (b :: S -> Type) (s :: S).
+        (PSubcategory f a, PSubcategory f b) =>
+        Term s (f a :--> (a :--> f b) :--> f b)
```
cjay, line 58: I think you'd want bind to be inlined most of the time (and inline its arguments into its body), so I suppose a Haskell function would be better.

cjay (approved):
I know almost nothing about category theory, so no idea what's going on in Star.hs, not sure if now is the right time for me to dig into that :)

src/Plutarch/Extra/Bind.hs
kozross: A Haskell function would actually _write_ better as well: `#>>=` versus `pbind` is not only more familiar-looking, but leads to much more readable code. To see how far this goes, just look at the laws! So I agree, definitely want to make that change.

peter-mlabs:
Is this good to merge now? (once the changelog is updated)?

kozross:
@peter-mlabs Should be. We can probably just roll this into 2.0.0.

peter-mlabs:
Changing the base to #48

PR #50, opened 2022-08-08
`mustCompile`: enable detail trace

chfanghr:


peter-mlabs:
`DetTracing` enables ["deterministic tracing"](https://github.com/Plutonomicon/plutarch-plutus/issues/532), not "detail". But it might still be more detailed, so it _might_ be what we want.

Do you have an example of where `DoTracing` vs `DetTracing` differ?

L-as:
You probably want this as I explained in the issue.

peter-mlabs (approved):
Merging

PR #51, opened 2022-08-08
Add extra compilation utilities

peter-mlabs:
These two modules have been factored out of the (newly created) `plutarch-benchmark` repository, which itself was factored out of LiqwidX. 

It includes utilities for precompilation of functions (which can make testing and benchmarking more efficient) as well as `DScript`s, which wraps a compiled script with both its `DetTracing` and `NoTracing` variants so that the `NoTracing` variant can be called, and libraries can lazily fall back to `DetTracing` upon error.

### DIFF
--- a/src/Plutarch/Extra/Compile.hs
+++ b/src/Plutarch/Extra/Compile.hs
@@ -2,9 +2,8 @@
 
 module Plutarch.Extra.Compile (mustCompile) where
 
-import Data.Default (def)
 import qualified Data.Text as T
-import Plutarch (ClosedTerm, compile)
+import Plutarch (ClosedTerm, Config (..), TracingMode (DetTracing), compile)
 import PlutusLedgerApi.V1 (Script)
 
 {- | Compile a ClosedTerm, throwing an error if unsuccessful.
@@ -12,6 +11,8 @@ import PlutusLedgerApi.V1 (Script)
      @since 2.0.0
 -}
 mustCompile :: ClosedTerm a -> Script
-mustCompile t = case compile def t of
+mustCompile t = case compile conf t of
     Left err -> error $ unwords ["Plutarch compilation error:", T.unpack err]
     Right s -> s
+  where
+    conf = Config{tracingMode = DetTracing}
--- a/src/Plutarch/Extra/Compile.hs
+++ b/src/Plutarch/Extra/Compile.hs
@@ -4,7 +4,7 @@ module Plutarch.Extra.Compile (mustCompile) where
 
 import qualified Data.Text as T
 import Plutarch (
-    Config (Config),
+    Config (Config, tracingMode),
     TracingMode (DetTracing),
     compile,
  )
--- a/src/Plutarch/Extra/Compile.hs
+++ b/src/Plutarch/Extra/Compile.hs
@@ -4,7 +4,7 @@ module Plutarch.Extra.Compile (mustCompile) where
 
 import qualified Data.Text as T
 import Plutarch (
-    Config (..),
+    Config (Config),
     TracingMode (DetTracing),
     compile,
  )
--- a/src/Plutarch/Extra/Compile.hs
+++ b/src/Plutarch/Extra/Compile.hs
@@ -2,6 +2,11 @@
 
 module Plutarch.Extra.Compile (mustCompile) where
 
+import Data.Default (Default (def))
+import qualified Data.Text as T
+import Plutarch (compile)
+import Plutarch.Prelude
+import PlutusLedgerApi.V2 (Script)
 import qualified Data.Text as T
 import Plutarch (Config (..), TracingMode (DetTracing), compile)
 import Plutarch.Prelude
--- a/src/Plutarch/Extra/Compile.hs
+++ b/src/Plutarch/Extra/Compile.hs
@@ -8,14 +8,13 @@ import Plutarch (
     TracingMode (DetTracing),
     compile,
  )
-import Plutarch.Prelude
 import PlutusLedgerApi.V2 (Script)
 
 {- | Compile a ClosedTerm, throwing an error if unsuccessful.
 
      @since 2.0.0
 -}
-mustCompile :: ClosedTerm a -> Script
+mustCompile :: forall (a :: S -> Type). ClosedTerm a -> Script
 mustCompile t = case compile conf t of
     Left err -> error $ unwords ["Plutarch compilation error:", T.unpack err]
     Right s -> s
--- a/src/Plutarch/Extra/Compile.hs
+++ b/src/Plutarch/Extra/Compile.hs
@@ -1,8 +1,16 @@
+{-# LANGUAGE RankNTypes #-}
+
 module Plutarch.Extra.Compile (mustCompile) where
 
-import Data.Default (Default (Data))
+import Data.Default (def)
+import qualified Data.Text as T
+import Plutarch (ClosedTerm, compile)
+import PlutusLedgerApi.V1 (Script)
 
--- | Compile a ClosedTerm, throwing an error if unsuccessful.
+{- | Compile a ClosedTerm, throwing an error if unsuccessful.
+
+     @since 2.0.0
+-}
 mustCompile :: ClosedTerm a -> Script
 mustCompile t = case compile def t of
     Left err -> error $ unwords ["Plutarch compilation error:", T.unpack err]
--- a/src/Plutarch/Extra/Compile.hs
+++ b/src/Plutarch/Extra/Compile.hs
@@ -3,8 +3,9 @@
 module Plutarch.Extra.Compile (mustCompile) where
 
 import qualified Data.Text as T
-import Plutarch (ClosedTerm, Config (..), TracingMode (DetTracing), compile)
-import PlutusLedgerApi.V1 (Script)
+import Plutarch (Config (..), TracingMode (DetTracing), compile)
+import Plutarch.Prelude
+import PlutusLedgerApi.V2 (Script)
 
 {- | Compile a ClosedTerm, throwing an error if unsuccessful.
 
--- a/src/Plutarch/Extra/Compile.hs
+++ b/src/Plutarch/Extra/Compile.hs
@@ -1,0 +1,9 @@
+module Plutarch.Extra.Compile (mustCompile) where
+
+import Data.Default (Default (Data))
+
+-- | Compile a ClosedTerm, throwing an error if unsuccessful.
+mustCompile :: ClosedTerm a -> Script
+mustCompile t = case compile def t of
+    Left err -> error $ unwords ["Plutarch compilation error:", T.unpack err]
+    Right s -> s
--- a/src/Plutarch/Extra/Compile.hs
+++ b/src/Plutarch/Extra/Compile.hs
@@ -2,13 +2,12 @@
 
 module Plutarch.Extra.Compile (mustCompile) where
 
-import Data.Default (Default (def))
 import qualified Data.Text as T
-import Plutarch (compile)
-import Plutarch.Prelude
-import PlutusLedgerApi.V2 (Script)
-import qualified Data.Text as T
-import Plutarch (Config (..), TracingMode (DetTracing), compile)
+import Plutarch (
+    Config (..),
+    TracingMode (DetTracing),
+    compile,
+ )
 import Plutarch.Prelude
 import PlutusLedgerApi.V2 (Script)
 
--- a/src/Plutarch/Extra/DScript.hs
+++ b/src/Plutarch/Extra/DScript.hs
@@ -21,7 +21,7 @@ import qualified Data.Text as Text
 import GHC.Generics (Generic)
 import Plutarch (
     Config (Config, tracingMode),
-    TracingMode (DoTracing, NoTracing),
+    TracingMode (DetTracing, NoTracing),
     compile,
  )
 import Plutarch.Evaluate (EvalError, evalScript)
@@ -50,7 +50,7 @@ checkedCompileD ::
     Either Text DScript
 checkedCompileD term = do
     script <- compile Config{tracingMode = NoTracing} term
-    debugScript <- compile Config{tracingMode = DoTracing} term
+    debugScript <- compile Config{tracingMode = DetTracing} term
     pure $ DScript{script, debugScript}
 
 -- | Like 'mustCompile', but with tracing turned on.
@@ -59,7 +59,7 @@ mustCompileTracing ::
     (forall (s :: S). Term s a) ->
     Script
 mustCompileTracing term =
-    case compile Config{tracingMode = DoTracing} term of
+    case compile Config{tracingMode = DetTracing} term of
         Left err -> error $ unwords ["Plutarch compilation error: ", T.unpack err]
         Right script -> script
 
--- a/src/Plutarch/Extra/DScript.hs
+++ b/src/Plutarch/Extra/DScript.hs
@@ -1,0 +1,196 @@
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE OverloadedRecordDot #-}
+{-# LANGUAGE RankNTypes #-}
+
+module Plutarch.Extra.DScript (
+    DScript (..),
+    checkedCompileD,
+    mustCompileD,
+    mustFinalEvalDScript,
+    finalEvalDScript,
+    mustEvalScript,
+    mustEvalD,
+) where
+
+import Control.DeepSeq (NFData)
+import Data.Text (Text)
+import qualified Data.Text as T
+import qualified Data.Text as Text
+import GHC.Generics (Generic)
+import Plutarch (
+    Config (Config, tracingMode),
+    TracingMode (DoTracing, NoTracing),
+    compile,
+ )
+import Plutarch.Evaluate (EvalError, evalScript)
+import Plutarch.Extra.Compile (mustCompile)
+import Plutarch.Prelude (S, Term, Type)
+import PlutusLedgerApi.V1 (ExBudget, Script)
+import UntypedPlutusCore.Evaluation.Machine.Cek (
+    CekUserError (CekEvaluationFailure, CekOutOfExError),
+    ErrorWithCause (ErrorWithCause),
+    EvaluationError (InternalEvaluationError, UserEvaluationError),
+ )
+
+-- | A 'Script' with a debug fallback that has tracing turned on.
+data DScript = DScript {script :: Script, debugScript :: Script}
+    deriving stock (Eq, Show, Generic)
+    deriving anyclass (NFData)
+
+{- | For handling compilation errors right away.
+
+ You pay for the compilation of the debug script, even if it's not needed down
+ the line. You most likely want 'mustCompileD' instead.
+-}
+checkedCompileD ::
+    forall (a :: S -> Type).
+    (forall (s :: S). Term s a) ->
+    Either Text DScript
+checkedCompileD term = do
+    script <- compile Config{tracingMode = NoTracing} term
+    debugScript <- compile Config{tracingMode = DoTracing} term
+    pure $ DScript{script, debugScript}
+
+-- | Like 'mustCompile', but with tracing turned on.
+mustCompileTracing ::
+    forall (a :: S -> Type).
+    (forall (s :: S). Term s a) ->
+    Script
+mustCompileTracing term =
+    case compile Config{tracingMode = DoTracing} term of
+        Left err -> error $ unwords ["Plutarch compilation error: ", T.unpack err]
+        Right script -> script
+
+{- | Compilation errors cause exceptions, but deferred by lazyness.
+
+ You don't pay for compilation of the debug script if it's not needed!
+-}
+mustCompileD ::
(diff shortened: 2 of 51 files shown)


src/Plutarch/Extra/Precompile.hs
```diff
@@ -0,0 +1,240 @@
...
+    forall p.
+    (HasCallStack, PLift p) =>
+    CompiledTerm p ->
+    PLifted p
+pliftCompiled prog = case pliftCompiled' prog of
+    Right x -> x
+    Left LiftError_FromRepr ->
+        error "plift failed: pconstantFromRepr returned 'Nothing'"
+    Left (LiftError_KnownTypeError e) ->
+        let unliftErrMaybe = e ^? _UnliftingErrorE
```
emiflake, line 233: Do we really need lens for this?
```hs
    Left (LiftError_KnownTypeError e) ->
       case e of
         KnownTypeUnliftingError unliftErr ->
           error $ "plift failed: incorrect type: " <> show unliftErr
         KnownTypeEvaluationFailure ->
           error $ "plift failed: absurd evaluation failure"
```

src/Plutarch/Extra/DScript.hs
```diff
@@ -0,0 +1,196 @@
...
+import Plutarch.Prelude (S, Term, Type)
+import PlutusLedgerApi.V1 (ExBudget, Script)
+import UntypedPlutusCore.Evaluation.Machine.Cek (
+    CekUserError (CekEvaluationFailure, CekOutOfExError),
+    ErrorWithCause (ErrorWithCause),
+    EvaluationError (InternalEvaluationError, UserEvaluationError),
+ )
+
+-- | A 'Script' with a debug fallback that has tracing turned on.
+data DScript = DScript {script :: Script, debugScript :: Script}
```
kozross, line 38: I guess the `D` refers to `debug` here? It'd be a bit clearer if it had that in its name, along with all its functions.

src/Plutarch/Extra/Precompile.hs
```diff
@@ -0,0 +1,240 @@
...
+--  Copied and adapted the stuff below from 'Plutarch.Lift'. Including
+--  'LiftError', because the data constructors aren't exported there. Also
+--  added trace messages in the exceptions.
+
+-- | Error during script evaluation.
+data LiftError
+    = LiftError_EvalError EvalError [Text]
+    | LiftError_KnownTypeError KnownTypeError
+    | LiftError_FromRepr
+    deriving stock (Eq)
```
kozross, line 205: The underscores in the names here aren't really idiomatic Haskell. Also, is something stopping this being a `Show` instance?

kozross (changes requested):
Needs a changelog entry, `since`s, and a minor version bump.

src/Plutarch/Extra/Precompile.hs
```diff
@@ -0,0 +1,240 @@
...
+
+applyDScript :: DScript -> DScript -> DScript
+applyDScript f a =
+    DScript
+        { script = applyScript f.script a.script
+        , debugScript = applyScript f.debugScript a.debugScript
+        }
+
+-- | Type-safe wrapper for compiled Plutarch functions.
+newtype CompiledTerm (a :: S -> Type) = CompiledTerm DScript
```
SeungheonOh, line 82: neat pick:
couldn't this be 
```hs
newtype CompiledTerm (a :: S -> Type) = CompiledTerm { toDScript :: DScript }
```

src/Plutarch/Extra/DScript.hs
SeungheonOh: I'd prefer `DebugScript`

src/Plutarch/Extra/Precompile.hs
peter-mlabs: This is lifted directly from [`Plutarch.Lift`](https://hoogle.nix.dance/file/nix/store/7r3zwxqgrvyd38z8myq5n9w7hzgnvj1m-plutarch-lib-plutarch-1.1.0-haddock-doc/share/doc/plutarch/html/src/Plutarch.Lift.html#LiftError). I've just asked in the discord if we can export these data constructors to simplify some of this code and the following functions.

peter-mlabs: @cjay may have more comments here

src/Plutarch/Extra/Precompile.hs
cjay: Yeah lens is not needed. My original intent was to stay as close as possible to the original code from Plutarch to have an easier time porting over changes, but the function is not really complex enough that it matters.

src/Plutarch/Extra/DScript.hs
cjay: I thought about that, wanted to avoid having it sound like a `DebugScript` is only for debugging. I'm not sure about this. Any name that better explains what's going on is too long, like `ScriptWithDebugFallback`.

src/Plutarch/Extra/Precompile.hs
cjay: Yeah I just tried to make minimal changes to the original code.

src/Plutarch/Extra/Precompile.hs
cjay: I'm disagree with this. `toDScript` is not a fitting name for a field, it's just abusing automatic generation of field selector functions. This makes using record-dot syntax as well as named field puns dealing with the field awkward. IMHO field selector functions should be deprecated anyway, as soon as record-dot syntax is solid.
Also, field selector functions on polymorphic fields (doesn't apply here) can have a type that has the `forall` in a bad place, potentially causing a need for eta-expansion (this was a hidden problem until simplified subsumption landed in GHC). For example, here the `forall` would be better of in the outer scope:
```
ghci> newtype Foo = Foo { ct :: forall (s :: S). Term s PInteger }
ghci> :t ct
ct :: Foo -> forall (s :: S). Term s PInteger
```

src/Plutarch/Extra/Precompile.hs
cjay: Btw, the original now includes a `LiftError_CompilationError`. We don't even want that, because the Script is already compiled.

src/Plutarch/Extra/Precompile.hs
SeungheonOh: Oh, I has some problems I was unaware of. Thanks for bringing this on.

src/Plutarch/Extra/DScript.hs
peter-mlabs: I'm digging through a thesaurus, and I'm coming up with 

- `ProvisionalScripts` (i.e., a script that is executed normally with the provisio that tracing is turned on in certain circumstances). I like this because "provisional" isn't a super over-loaded term, so most people won't come away with misunderstanding just from a glance
- "ContingencyScripts" 
- "ConditionalScripts" -- but this one uses the (very) overloaded "Conditional".

src/Plutarch/Extra/Precompile.hs
peter-mlabs: I've gotten the OK to make a PR to export these

peter-mlabs: > Is something stopping this being a `Show` instance?

The wrapped PlutusCore type doesn't have a show instance.

SeungheonOh:
Another way we can skip all the code duplication from `plift` code is to just use Plutarch `plift` itself. There is `RCompiled UTerm` in `Plutarch.Internal`, so I think we can add functionality to make a make Plutarch Term from `DScript`. That way we can just use `plift` from Plutarch itself

src/Plutarch/Extra/DScript.hs
kozross: How about `DebuggableScript`?

src/Plutarch/Extra/Precompile.hs
peter-mlabs: https://github.com/Plutonomicon/plutarch-plutus/pull/535

cjay:
Do we *really* want `DetTracing`? Las said `DoTracing` includes call stacks, while `DetTracing` doesn't (though I don't think I've seen a call stack while doing manual testing, weird).

cjay:
@SeungheonOh interesting find. Though the official `plift` doesn't output trace messages at all. Maybe we should try to get a change into the official one.

SeungheonOh:
we'd want `tryplift :: Term s a -> Either Trace (PLift a)`.

peter-mlabs:
> Do we _really_ want `DetTracing`? Las said `DoTracing` includes call stacks, while `DetTracing` doesn't (though I don't think I've seen a call stack while doing manual testing, weird).

Oh, I don't know. My move to `DetTracing`came from [this](https://github.com/Plutonomicon/plutarch-plutus/issues/532#issuecomment-1208135958) comment.

SeungheonOh:
https://github.com/Liqwid-Labs/liqwid-plutarch-extra/commit/5349c1d78264860fb514d00f7505f98a687acc03 I tried this.
```
λ d = plam $ \(x :: Term s PInteger) -> x + 10
λ e = compile' d
λ f = applyCompiledTerm e (pconstant 10)
λ g = toPTerm f
λ plift g
20
```
it seems to work though this is a very quick and dirty test.


I also attempted to make `DScript (Config -> Either Text Script)` but it seems like a quite big change and need some opinion before I go a head and make the changes.
@cjay @peter-mlabs

cjay:
@SeungheonOh nice.
Hmm `DScript (Config -> Either Text Script)` is an interesting idea. We'd need to make sure the resulting Script gets memoized, so that it doesn't need to be compiled every time. I think it can be done by creating a lookup table lazily, and having the function inside the `DScript` just be a lookup into that table. Or we just store the lookup table in `DScript` directly, that's more straight-forward and flexible :)

kozross:
@cjay There's a library that might make this easier [here](https://hackage.haskell.org/package/monad-memo).

SeungheonOh:
Plutonomicon/plutarch-plutus#536

cjay:
Current version can't work because my `LiftError` had trace messages, the official one doesn't.

peter-mlabs:
So where is this at now? I suppose I was overzealous with the changes. Do we need the original back?

SeungheonOh:
I don't think we are going to make any significant change besides making it use `Plutarch.plift` instead of having it's own.

peter-mlabs:
I'm confused. 

- > Current version can't work because @cjay 's `LiftError` had trace messages
- [plutarch-plutus/#536](https://github.com/Plutonomicon/plutarch-plutus/pull/536) was going to give us traces for lifting, but got closed in favor of [plutarch-plutus/#538](https://github.com/Plutonomicon/plutarch-plutus/pull/538)
- [plutarch-plutus/#538](https://github.com/Plutonomicon/plutarch-plutus/pull/538) is now closed as well.

cjay:
The one with `evalTerm` that got merged is https://github.com/Plutonomicon/plutarch-plutus/pull/539. Looks like we have to put the `Script` into the `Term`, call `evalTerm` to catch trace messages, and then `plift`, which will compile and evaluate again. TBH I have no idea why this is considered better than just having the official `plift'` support trace messages. I'll make the changes.

SeungheonOh:
Well "compile" doesn't happen when you give term with Raw UPLC in it. Las was not in favor of changing `plift'` because `evalTerm` covers that, and `evalTerm` looked more useful. 

He talked about making lens between `CompiledTerm` and `ClosedTerm` would be helpful.

cjay:
Changes are in. `evalTerm` wasn't even needed, since `DebuggableScript` needs to do evaluation anyway, to do the debug fallback.
I don't think we gained anything in going this way, at all. Script user errors are now exceptions. If someone wants to analyze errors (in case of testing and expecting errors), that sucks. One can use `pliftCompiled'` to get the `LiftError` and analyze / match on that, but to check if the trace messages fit the expected error, one has to catch exceptions now.

cjay:
If errors are expected, execution could be done in a different way, without trying to lift. It still might complicate things.

cjay:
I guess I could package up the `(Either EvalError Script, ExBudget, [Text])` together with the `LiftError`. Ugly, but would give all the info.

peter-mlabs:
@cjay fwiw I dont mind if you'd prefer going back to the original setup, or have a different idea.

peter-mlabs:
Oops, the merge in [cd94540](https://github.com/Liqwid-Labs/liqwid-plutarch-extra/pull/51/commits/cd945407f0d371aa8e1f3d4937d5ff7ab96a96f3) accidentally reintroduced a test suite. 

I've just removed it again, and once CI passes we can merge this. @cjay, if you want to make any additional modification feel free.

src/Plutarch/Extra/Compile.hs
```diff
@@ -3,8 +3,13 @@
 module Plutarch.Extra.Compile (mustCompile) where
 
 import qualified Data.Text as T
-import Plutarch (ClosedTerm, Config (..), TracingMode (DetTracing), compile)
-import PlutusLedgerApi.V1 (Script)
+import Plutarch (
+    Config (..),
```
kozross, line 7: Wildcard import.

src/Plutarch/Extra/DebuggableScript.hs
```diff
@@ -0,0 +1,196 @@
...
+import Plutarch.Extra.Compile (mustCompile)
+import Plutarch.Prelude (S, Term, Type)
+import PlutusLedgerApi.V1 (ExBudget, Script)
+import UntypedPlutusCore.Evaluation.Machine.Cek (
+    CekUserError (CekEvaluationFailure, CekOutOfExError),
+    ErrorWithCause (ErrorWithCause),
+    EvaluationError (InternalEvaluationError, UserEvaluationError),
+ )
+
+-- | A 'Script' with a debug fallback that has tracing turned on.
```
kozross, line 37: sinces.

src/Plutarch/Extra/DebuggableScript.hs
```diff
@@ -0,0 +1,196 @@
...
+data DebuggableScript = DebuggableScript {script :: Script, debugScript :: Script}
+    deriving stock (Eq, Show, Generic)
+    deriving anyclass (NFData)
+
+{- | For handling compilation errors right away.
+
+ You pay for the compilation of the debug script, even if it's not needed down
+ the line. You most likely want 'mustCompileD' instead.
+-}
```
kozross, line 46: since.

src/Plutarch/Extra/DebuggableScript.hs
```diff
@@ -0,0 +1,196 @@
...
+checkedCompileD ::
+    forall (a :: S -> Type).
+    (forall (s :: S). Term s a) ->
+    Either Text DebuggableScript
+checkedCompileD term = do
+    script <- compile Config{tracingMode = NoTracing} term
+    debugScript <- compile Config{tracingMode = DetTracing} term
+    pure $ DebuggableScript{script, debugScript}
+
+-- | Like 'mustCompile', but with tracing turned on.
```
kozross, line 56: This isn't exported, so regular comment, not Haddock.

src/Plutarch/Extra/DebuggableScript.hs
```diff
@@ -0,0 +1,196 @@
...
+    Script
+mustCompileTracing term =
+    case compile Config{tracingMode = DetTracing} term of
+        Left err -> error $ unwords ["Plutarch compilation error: ", T.unpack err]
+        Right script -> script
+
+{- | Compilation errors cause exceptions, but deferred by lazyness.
+
+ You don't pay for compilation of the debug script if it's not needed!
+-}
```
kozross, line 69: since

src/Plutarch/Extra/Precompile.hs
```diff
@@ -0,0 +1,218 @@
...
+import GHC.Stack (HasCallStack)
+import Plutarch.Extra.DebuggableScript (
+    DebuggableScript (DebuggableScript),
+    debugScript,
+    mustCompileD,
+    mustEvalD,
+    mustFinalEvalDebuggableScript,
+    script,
+ )
+import Plutarch.Internal (RawTerm (RCompiled), Term (..), TermResult (TermResult))
```
kozross, line 43: Wildcard import.

src/Plutarch/Extra/Precompile.hs
```diff
@@ -0,0 +1,218 @@
+{-# LANGUAGE OverloadedRecordDot #-}
+{-# LANGUAGE RankNTypes #-}
+{-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
+
+{-# HLINT ignore "Use camelCase" #-}
```
kozross, line 5: Explanatory note would be good.

src/Plutarch/Extra/Precompile.hs
```diff
@@ -0,0 +1,218 @@
...
+
+{- | Pre-compiling Plutarch functions and applying them.
+
+ Speeds up benchmarking and testing.
+-}
+module Plutarch.Extra.Precompile (
```
kozross, line 11: sinces

kozross (changes requested):

peter-mlabs:
Will fix. Thanks Koz

src/Plutarch/Extra/Compile.hs
peter-mlabs: This will actually be deleted once the Prelude PR is merged.

src/Plutarch/Extra/Precompile.hs
peter-mlabs: This actually can be removed, I think. I think it had to do with the `LiftError`, which had underscores (because it was mimicking the upstream)

src/Plutarch/Extra/DebuggableScript.hs
```diff
@@ -0,0 +1,211 @@
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE OverloadedRecordDot #-}
+{-# LANGUAGE RankNTypes #-}
+
+-- | @since 3.0.2
```
kozross, line 7: Modules themselves don't usually have `since`s: my comment was a shorthand for 'everything exported in here'.

src/Plutarch/Extra/DebuggableScript.hs
```diff
@@ -0,0 +1,211 @@
...
+    ErrorWithCause (ErrorWithCause),
+    EvaluationError (InternalEvaluationError, UserEvaluationError),
+ )
+
+{- | A 'Script' with a debug fallback that has tracing turned on.
+ @since 3.0.2
+-}
+data DebuggableScript = DebuggableScript {script :: Script, debugScript :: Script}
+    deriving stock (Eq, Show, Generic)
+    deriving anyclass (NFData)
```
kozross, line 43: These need `since`s as well.

src/Plutarch/Extra/Precompile.hs
```diff
@@ -0,0 +1,242 @@
+{-# LANGUAGE OverloadedRecordDot #-}
+{-# LANGUAGE RankNTypes #-}
+
+{- | Pre-compiling Plutarch functions and applying them.
+
+ Speeds up benchmarking and testing.
+
+ @since 3.0.2
```
kozross, line 8: Again, that was a shorthand for '`since`s on everything public in here'.

src/Plutarch/Extra/Precompile.hs
```diff
@@ -0,0 +1,242 @@
...
+compile' t = CompiledTerm $ mustCompileD t
+
+{- | Convert a 'CompiledTerm' to a 'Script'.
+ @since 3.0.2
+-}
+toDebuggableScript :: forall (a :: S -> Type). CompiledTerm a -> DebuggableScript
+toDebuggableScript (CompiledTerm dscript) = dscript
+
+-- | @since 3.0.2
+toEvaluatedTerm :: CompiledTerm a -> (forall (s :: S). Term s a)
```
kozross, line 107: Kind signature and explicit `forall` on `a`.

src/Plutarch/Extra/Precompile.hs
```diff
@@ -0,0 +1,242 @@
...
+infixl 7 ###~
+
+{- | Convert a 'CompiledTerm' to the associated Haskell value. Fail otherwise.
+
+ This will fully evaluate the compiled term, and convert the resulting value.
+
+ @since 3.0.2
+-}
+pliftCompiled' ::
+    forall p. PUnsafeLiftDecl p => CompiledTerm p -> Either LiftError (PLifted p)
```
kozross, line 231: Kind signature on `p`.

src/Plutarch/Extra/Precompile.hs
```diff
@@ -0,0 +1,242 @@
...
+pliftCompiled' ct = plift' def $ toEvaluatedTerm ct
+
+{- | Like `pliftCompiled'` but throws on failure.
+ @since 3.0.2
+-}
+pliftCompiled ::
+    forall p.
```
kozross, line 238: Kind signature on `p`.

kozross (changes requested):

src/Plutarch/Extra/DebuggableScript.hs
peter-mlabs: Ah. Also need a line break between some @sinces...

kozross (approved):

PR #52, opened 2022-08-09
`CompiledTerm` to Plutarch `ClosedTerm`

SeungheonOh:
I think `toPTerm` needs a better name.

#51

### DIFF
--- a/src/Plutarch/Extra/Precompile.hs
+++ b/src/Plutarch/Extra/Precompile.hs
@@ -45,18 +45,19 @@ import Plutarch.Extra.DScript (
     mustEvalD,
     script,
  )
+import Plutarch.FFI (unsafeForeignImport)
 import Plutarch.Lift (
     PConstantDecl (pconstantFromRepr),
     PUnsafeLiftDecl (PLifted),
  )
-import Plutarch.Prelude (PLift, S, Term, Type, (:-->), ClosedTerm)
-import Plutarch.Internal (Term (Term), TermResult(TermResult), RawTerm(RCompiled))
+import Plutarch.Prelude (ClosedTerm, PLift, S, Term, Type, (:-->))
 import PlutusCore.Builtin (KnownTypeError, readKnownConstant)
 import PlutusCore.Evaluation.Machine.Exception (_UnliftingErrorE)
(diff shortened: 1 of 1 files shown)


src/Plutarch/Extra/Precompile.hs
```diff
@@ -48,12 +49,14 @@ import Plutarch.Lift (
     PConstantDecl (pconstantFromRepr),
     PUnsafeLiftDecl (PLifted),
  )
-import Plutarch.Prelude (PLift, S, Term, Type, (:-->))
+import Plutarch.Prelude (PLift, S, Term, Type, (:-->), ClosedTerm)
+import Plutarch.Internal (Term (Term), TermResult(TermResult), RawTerm(RCompiled))
```
L-as, line 53: really big no IMO... what are you trying to do?

L-as:
I see the point, but you want to use Plutarch.FFI here.

SeungheonOh:
`unsafeForeignImport` ?

SeungheonOh:
Okay, It's not using `Plutarch.FFI`. It's okay to use `unsafeForeignImport` because `CompiledTerm` will always have correct Plutarch type.

src/Plutarch/Extra/Precompile.hs
```diff
@@ -196,13 +197,17 @@ infixl 7 ###
...
+ This can then be used to `plift`.
 -}
-toPTerm :: forall p. CompiledTerm p -> ClosedTerm p
-toPTerm prog = Term $ const $ pure $ TermResult (RCompiled $ UPLC._progTerm $ unScript . debugScript . toDScript $ prog) []
+toPTerm ::
+    forall (p :: S -> Type).
+    CompiledTerm p ->
+    ClosedTerm p
+toPTerm prog =
+    let (Script (UPLC.Program _ version term)) = debugScript . toDScript $ prog
```
cjay, line 208: Why does this use `debugScript`? Pretty sure it should use `finalEvalDScript`, like `pliftCompiled'`.

cjay: Hmm and unless we can get the official `plift` to output trace messages, maybe it should only use `script`. I'll ask if we could get that change in.

PR #57, opened 2022-08-12
use `Plutarch.Prelude` Mixin

SeungheonOh:


chfanghr (approved):
🚀

peter-mlabs:
@kozross, how do you feel about allowing `import Plutarch.Prelude` without an explicit import list, or the way we do for dApps with `mixins` and an implicit prelude?

kozross:
I feel that `Plutatch.Prelude` is not a full replacement for the `base` one, and therefore should be treated like any other import.

If we want a replacement `Prelude`, which may or may not include identifiers coming from Plutarch, that's a separate story IMHO, since there's definitely some disagreement over what it should, and shouldn't, include. I certainly have some thoughts of my own, but they'd need a full writeup.

peter-mlabs:
FTR, in all the non Liqwid Lib's repos (as well as [plutus-scaffold](https://github.com/mlabs-haskell/plutus-scaffold/blob/master/onchain/template-project.cabal), we use

```
  mixins:
    base hiding (Prelude),
    pprelude (PPrelude as Prelude)
``` 

in the  cabal and

```hs
module PPrelude (
  module Prelude,
  module Plutarch.Prelude,
) where

import Plutarch.Prelude
import Prelude
```

as `PPrelude`. I think this would be good to adopt if there's no strong opposition.

kozross:
@peter-mlabs I'm not opposed - there are no name clashes or major surprises, and the Plutarch prelude is its intended (primary) API.

SeungheonOh:
Okay, It seems like we are going to use `PPrelude` I'll update this PR.

src/Plutarch/Extra/FixedDecimal.hs
```diff
@@ -23,9 +23,7 @@ import GHC.TypeLits (KnownNat, Nat, natVal)
 import qualified Generics.SOP as SOP
 import Plutarch.Api.V1 (AmountGuarantees, KeyGuarantees, PValue)
 import Plutarch.Api.V1.Value (psingletonValue)
-import Plutarch.Bool (PEq, POrd, PPartialOrd)
 import Plutarch.Extra.Function (pflip)
-import Plutarch.Integer (PInteger, PIntegral (pdiv))
 import Plutarch.Num (PNum (..))
```
kozross, line 27: A bit stale, but proper imports.

src/Plutarch/Extra/FixedDecimal.hs
```diff
@@ -23,9 +23,7 @@ import GHC.TypeLits (KnownNat, Nat, natVal)
...
 import qualified Plutarch.Numeric.Additive as A (
     AdditiveMonoid (..),
```
kozross, line 29: Also here.

src/Plutarch/Extra/FixedDecimal.hs
```diff
@@ -35,7 +33,12 @@ import Plutarch.Prelude (
     DerivePlutusType (..),
```
kozross, line 33: And here.

src/Plutarch/Extra/List.hs
```diff
@@ -24,33 +24,49 @@ module Plutarch.Extra.List (
...
+import Plutarch.Extra.TermCont (pletC)
+import Plutarch.Prelude (
+    PAsData,
+    PBool (..),
+    PBuiltinPair,
+    PEq (..),
+    PInteger,
+    PIsData,
+    PIsListLike,
+    PListLike (..),
```
kozross, line 41: Proper imports please.

src/Plutarch/Extra/Map/Sorted.hs
```diff
@@ -4,35 +4,38 @@
...
+import Plutarch.Extra.Map (pkeys)
+import Plutarch.Extra.TermCont (pletC)
+import Plutarch.Prelude (
+    PAsData,
+    PBool,
+    PBuiltinList,
+    PBuiltinPair,
+    PEq ((#==)),
+    PIsData,
+    PListLike (..),
```
kozross, line 18: Proper imports.

src/Plutarch/Extra/These.hs
```diff
@@ -42,12 +33,25 @@ import Plutarch.Extra.Traversable (PTraversable (ptraverse, ptraverse_))
...
     DPTStrat,
     DerivePlutusType,
+    PAsData,
+    PBool (PFalse, PTrue),
+    PDataRecord,
+    PEq,
+    PIsData,
+    PLabeledType ((:=)),
+    POrd,
+    PPartialOrd (..),
```
kozross, line 43: Proper imports.

kozross (changes requested):

SeungheonOh:
Okay using mix-ins looks so much cleaner.

peter-mlabs:
Added `-Wmissing-import-lists` to the cabal and `{-# OPTIONS_GHC -Wno-missing-import-lists #-}` to the files with `import Prelude hiding (..,)` statements.

peter-mlabs (approved):
All should be good now, thanks!

src/Plutarch/Extra/AssetClass.hs
```diff
@@ -3,6 +3,7 @@
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -Wno-missing-import-lists #-}
```
kozross, line 6: A comment explaining why this is off would be good I think, for future reference if nothing else.

src/Plutarch/Extra/Bind.hs
```diff
@@ -1,6 +1,7 @@
 {-# LANGUAGE QuantifiedConstraints #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE UndecidableInstances #-}
+{-# OPTIONS_GHC -Wno-missing-import-lists #-}
```
kozross, line 4: Same here.

src/Plutarch/Extra/Map.hs
```diff
@@ -1,6 +1,7 @@
 {-# LANGUAGE PartialTypeSignatures #-}
 {-# LANGUAGE QuantifiedConstraints #-}
 {-# LANGUAGE ViewPatterns #-}
+{-# OPTIONS_GHC -Wno-missing-import-lists #-}
```
kozross, line 4: Also here.

src/Plutarch/Extra/Traversable.hs
```diff
@@ -1,5 +1,6 @@
 {-# LANGUAGE QuantifiedConstraints #-}
 {-# LANGUAGE TypeApplications #-}
+{-# OPTIONS_GHC -Wno-missing-import-lists #-}
```
kozross, line 3: And here.

test/Spec.hs
```diff
@@ -5,7 +5,8 @@ import Test.Tasty (defaultMain, testGroup)
 
 import qualified Spec.Extra.List as List
 import qualified Spec.Extra.Map as Map
-import qualified Spec.Extra.MultiSig as MultiSig
+
+-- import qualified Spec.Extra.MultiSig as MultiSig
```
kozross, line 9: Should this be removed?

test/Spec/Extra/Map/Sorted.hs
```diff
@@ -26,11 +26,7 @@ import qualified GHC.Generics as GHC
 
 --------------------------------------------------------------------------------
 
-import Plutarch (Term, phoistAcyclic, plam, (#), type (:-->))
-import Plutarch.Api.V2.AssocMap (KeyGuarantees (..), PMap)
-import Plutarch.Bool (PBool)
-import Plutarch.Builtin (PBuiltinPair, pfstBuiltin, psndBuiltin)
-import Plutarch.Integer (PInteger)
+import Plutarch.Api.V1.AssocMap (KeyGuarantees (..), PMap)
```
kozross, line 29: Wildcard imports shouldn't be used.

kozross (changes requested):

src/Plutarch/Extra/FixedDecimal.hs
```diff
@@ -18,39 +18,16 @@ module Plutarch.Extra.FixedDecimal (
...
-import Plutarch.Api.V2 (AmountGuarantees, KeyGuarantees)
-import Plutarch.Bool (PEq, POrd, PPartialOrd)
+import Plutarch.Api.V1 (AmountGuarantees, KeyGuarantees, PValue)
 import Plutarch.Extra.Function (pflip)
 import Plutarch.Extra.Value (psingletonValue)
-import Plutarch.Integer (PInteger, PIntegral (pdiv))
-import Plutarch.Num (PNum (..))
+import Plutarch.Num (PNum (pfromInteger, (#*)))
 import qualified Plutarch.Numeric.Additive as A (
     AdditiveMonoid (..),
```
kozross, line 28: Imports.

src/Plutarch/Extra/FixedDecimal.hs
```diff
@@ -18,39 +18,16 @@ module Plutarch.Extra.FixedDecimal (
...
     AdditiveSemigroup (..),
```
kozross, line 29: Also here.

src/Plutarch/Extra/Identity.hs
```diff
@@ -7,25 +7,8 @@ module Plutarch.Extra.Identity (
     PIdentity (..),
```
kozross, line 7: Imports.

kozross (changes requested):

kozross (approved):

PR #58, opened 2022-08-12
Explicit control over evaluated terms

SeungheonOh:
I think this can replace `CompiledTerm a` with `EvaluatedTerm 'False a`. 

Any name suggestion to replace `EvaluatedTerm` would be very helpful. (I don't like it's current name).

I will add docstrings once this is a usable idea.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2943,16 +2943,16 @@
         "secp256k1-haskell": "secp256k1-haskell"
       },
       "locked": {
-        "lastModified": 1659381657,
-        "narHash": "sha256-W0EWY0DV2idbIeqtJnShVHQ83exME8TK2GYQGbVpz8A=",
+        "lastModified": 1660245357,
+        "narHash": "sha256-Q93Lq6leerCQedmN+3lJQWotyuGAhW6Sn14EKZd2SsA=",
         "owner": "Plutonomicon",
         "repo": "plutarch-plutus",
-        "rev": "45b7c77a9ee9bd6c7dc25ddebcc3d12c58c4c3a2",
+        "rev": "3fe25c2376bca1a563ceecc273ae31771264e089",
         "type": "github"
       },
       "original": {
         "owner": "Plutonomicon",
-        "ref": "staging",
+        "ref": "master",
         "repo": "plutarch-plutus",
         "type": "github"
       }
--- a/flake.nix
+++ b/flake.nix
@@ -15,7 +15,7 @@
 
     # Plutarch and its friends
     plutarch = {
-      url = "github:Plutonomicon/plutarch-plutus?ref=staging";
+      url = "github:Plutonomicon/plutarch-plutus?ref=master";
 
       inputs.emanote.follows =
         "plutarch/haskell-nix/nixpkgs-unstable";
--- a/liqwid-plutarch-extra.cabal
+++ b/liqwid-plutarch-extra.cabal
@@ -100,6 +100,7 @@ library
     Plutarch.Extra.TermCont
     Plutarch.Extra.These
     Plutarch.Extra.Traversable
+    Plutarch.Extra.Evaluated
     Plutarch.Orphans
 
   build-depends:
@@ -108,6 +109,7 @@ library
     , generics-sop
     , plutarch-extra
     , plutarch-numeric
+    , plutus-core
     , tagged
     , text
 
--- a/src/Plutarch/Extra/Evaluated.hs
+++ b/src/Plutarch/Extra/Evaluated.hs
@@ -1,0 +1,108 @@
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ImpredicativeTypes #-}
+{-# LANGUAGE GADTs #-}
+{-# OPTIONS_GHC -Wno-all #-}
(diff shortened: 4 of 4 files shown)


cjay:
Can you clarify what pros and cons you see in this approach, and why it needs to happen? The existing stuff all works well and was carefully designed, so we need good reasons to throw it out and do it all over again.

Having compiled stuff inside `Terms` is nice, and using Terms instead of `CompiledTerm` directly would have some advantages (no extra operators) and some disadvantages (a lot of function calling around the various arguments to selectively pre-evaluate, and less type safety around having things pre-compiled). However, that's not what we're doing here.

If we wrap things up in our own type anyway, the internal representation doesn't seem that important to me. Having a compiled `Script` sit inside a `Term` is an additional indirection. Why do we need that?

Do you have a plan for how to do debug fallbacks properly? Because if you pre-evaluate an argument, that's a compiled `Script` sitting somewhere inside your wrapper type. If you try to compile that thing again, with `DoTracing` on, you're probably not getting your trace messages back.

The `Bool` is a case of boolean blindness, it should probably instead be `data CompiledTermState = Evaluated | NonEvaluated` or something similar.

The obvious name is still `CompiledTerm`, though if you want to use it to represent normal `Term`s too (since it can), I'd go with `Term'` or `GenericTerm`. In that case the tag should represent compiledness too, like `data CompiledTermState = Uncompiled | Compiled | Evaluated`.

With the choice of evaluation not part of the operators, that's going to require a lot of extra calls to `evaluateTerm` when benchmarking, it wouldn't be any less cumbersome than using `Term`s directly. That's why I added the operators with the `~` suffix.

I'm also not sure about how to even take advantage of having evaluatedness in the type in benchmarking code. `plutarch-benchmark` currently leaves the part of applying the compiled function to arguments to the user. Changing it to be automatic is messy, since funcitons can have different arities, and require input conversions. If that stays the same, the extra type info isn't going to change how likely users are to mess up.

SeungheonOh:
In terms of testing and bench marking, I feel like the notion of being compiled/non-compiled is less important than the notion of being evaluated/non-evaluated. Having a compiled Script and applying that doesn't have any significant performance benefit as oppose to having an evaluated script and applying it. `EvaluatedTerm` attempts to only take notice on if something is evaluated or not. It's only purpose is to ensure a term gets evaluated to be used in tests/bench-marks. That was not possible with `CompiledTerm` because some compiled terms are evaluated and some are not. 

`EvaluatedTerm` merely wraps around regular `Term` because `Term` provides all functionalities we need. While having `Script` can be more straightforward, that require dealing raw UPLC application and running CEK machine. I wanted to put all of those very internal functionalities into Plutarch and use them via Plutarch. 

Debugging is the part I'm more concerned with. Because `EvaluatedTerm` is a simple wrapper around `Term` with type safety, it wouldn't have a way to do what `CompiledTerm` can in terms of tracing. It's partly because this is mainly intended to be used in PQ as well. Any idea would be very helpful.

Boolean blindness will be fixed. Thanks for noticing me.

cjay:
Precompiling didn't have a big speedup for small functions, but it wasn't neglegible either, when I measured. I wouldn't be surprised if it makes a significant difference on big Scripts. I discovered that Plutarch can have exploding compile times if branching functions are nested deeply. A sorting network of size 8 takes about 2 minutes to compile, though admittedly that is highly unusual Plutarch code. Testing/benchmarking is already slow, so I care about saving some time.

Why do you care about evaluatedness for testing? Is it for checking for equality without having a `PEq` instance?

> That was not possible with CompiledTerm because some compiled terms are evaluated and some are not.

What exactly is not possible? Precise control of evaluation is possible, it's just not tracked in the types. It's nice to have it in the types and i'm not opposed to adding it there. Though I don't know how you plan to take advantage of this type level info. I think the type level tag for evaluatedness could just be added to the existing `CompiledTerm` implementation. 

Raw ULPC application was trivial, and is not exposed to users of the library, so I don't see any problem with that.
I don't remember interacting with CEK machine internals either. It's not a secret that we are dealing with Scripts, so there is no problem in needing to run a Script, instead of calling something that only operates on Terms.

Not needing a copy of `plift` would be nice, but can't that be accomplished without basing everything on `Term`s only? Plutarch should just offer an official function to put a `Script` into a `Term`, if using `RCompiled` directly is discouraged. It seems to be an official feature (as opposed to an implementation detail) that `Term`s can carry `Script`s, since the `evalTerm` function you got into Plutarch wouldn't make sense without this being official. 
It would also make sense to have such a function to be able to use Scripts written in other languages from Plutarch, as a kind of "FFI import". Though I suppose Plutarch doesn't have an "ABI".

SeungheonOh:
Ensuring something is evaluated or not is crucial more so in property testing. It allows us to shrink efficiently, without getting every computations copied over and over and over again, exploding CEK machine at the end. Also, while I wasn't aware of the fact that you can check without `PEq` using `UPLC`, it would make some instance of testing much easier to work with. I just wanted to remove interaction with `UPLC` as much as possible. However, it seems like it's not really a problem for `CompiledTerm`. 

I'll either add independent wrapper in PQ or find some other ways. Thanks

PR #63, opened 2022-08-16
Define Plut, replace uses of Top

kozross:
This eliminates (most) uses of `generics-sop`. In light of this, we also vendor a version of `Top` (named `Plut`). This also allows us to generalize `.=` to not need `SListI` anymore.

One module `Plutus.Extra.IsData` still uses SOP generics in a fundamental way. As replacing this would be significant, this PR does not address this issue.

Furthermore, in light of the upstream Plutarch changes, we now globally enable `DeriveAnyClass`, `DeriveGeneric` and `TypeFamilies`; the standards document has been updated with this in mind.


SeungheonOh (approved):
LGTM.

peter-mlabs (approved):

PR #66, opened 2022-08-16
pliftCompiled: Fallback on LiftErrors, trace messages

cjay:
This was very painful to write. Would be a lot less complicated without using the official `plift'`, but the fallback in case of successfull evaluation followed by a `LiftError` wasn't in the original implementation.


cjay:
No idea why the `hie.yaml` fixes HLS, but it does.

cjay:
done

peter-mlabs (approved):

PR #70, opened 2022-08-22
Make `MultiSig` more generic

chfanghr:


SeungheonOh (approved):
This seems like a good upgrade for `MultiSig`! Can you change the PR name to address that `MultiSig` is now generic? It's more than a efficiency improvement of it. 

LGTM.

src/Plutarch/Extra/MultiSig.hs
```diff
@@ -119,13 +95,53 @@ deriving via
...
+    r ->
+    Term s PMultiSigContext
+pmultiSigContext f = pcon (PMultiSigContext f.signatories)
+
 {- | Check if a Haskell-level MultiSig signs this transaction.
 
      @since 0.1.0
 -}
-validatedByMultisig :: forall (s :: S). MultiSig -> Term s (PTxInfo :--> PBool)
+validatedByMultisig :: forall (s :: S). MultiSig -> Term s (PMultiSigContext :--> PBool)
```
kozross, line 144: Signature change, so the `since` should be bumped too.

src/Plutarch/Extra/MultiSig.hs
```diff
@@ -134,18 +150,16 @@ validatedByMultisig params =
 
      @since 0.1.0
 -}
-pvalidatedByMultisig :: forall (s :: S). Term s (PMultiSig :--> PTxInfo :--> PBool)
+pvalidatedByMultisig :: forall (s :: S). Term s (PMultiSig :--> PMultiSigContext :--> PBool)
```
kozross, line 153: Same here.

kozross (changes requested):

kozross (approved):

PR #72, opened 2022-08-23
Clean up Extra.Map

kozross:
This resolves a range of issues with `Plutarch.Extra.Map.*`, mostly to do with correctness. This also adds clearer documentation, and improves the performance of several functions that were exported previously by avoiding intermediate copies.


src/Plutarch/Extra/Map.hs
```diff
@@ -1,185 +1,266 @@
...
+
+ @since 1.1.0
+-}
+pkeysEqual ::
+    forall (k :: S -> Type) (a :: S -> Type) (b :: S -> Type) (s :: S).
+    (PIsData k, PEq k) =>
+    Term s (PMap 'Sorted k a :--> PMap 'Sorted k b :--> PBool)
+pkeysEqual = phoistAcyclic $
+    plam $ \kvs kvs' ->
+        pmatch kvs $ \case
```
SeungheonOh, line 63: we don't need a lambda case here.

src/Plutarch/Extra/Map.hs
```diff
@@ -1,185 +1,266 @@
...
+        (v :: S -> Type)
+        (keys :: KeyGuarantees)
+        (s :: S).
+    (PListLike ell, PElemConstraint ell (PAsData k)) =>
+    Term s (PMap keys k v :--> ell (PAsData k))
 pkeys = phoistAcyclic $
-    plam $ \m -> unTermCont $ do
-        PMap kvs <- pmatchC m
-        pure $ pfmap # pfstBuiltin # kvs
+    plam $ \kvs -> pmatch kvs $ \case
```
SeungheonOh, line 210: we don't need a lambda case here.

src/Plutarch/Extra/Map.hs
```diff
@@ -1,185 +1,266 @@
...
+ Performs linearly with respect to its argument.
+
+ = Note
+
+ If there are duplicate keys in the input, the /last/ key will \'win\' in a
+ lookup.
+
+ @since 3.2.0
+-}
+pmapFromFoldableUnsorted ::
```
SeungheonOh, line 153: `punsortedMapFromFoldable` would be a better name in my opinion. same goes with sorted version.

src/Plutarch/Extra/Map.hs
```diff
@@ -1,185 +1,266 @@
...
+    go ::
+        forall (s' :: S).
+        Term s' (PBuiltinList (PBuiltinPair (PAsData k) (PAsData v))) ->
+        (Term s' k, Term s' v) ->
+        Term s' (PBuiltinList (PBuiltinPair (PAsData k) (PAsData v)))
+    go acc (key, val) =
+        pcon . PCons (ppairDataBuiltin # pdata key # pdata val) $ acc
 
-{- | Get the keys of a given map, the order of the keys is preserved.
+{- | Given a 'Foldable' of (not necessarily sorted) key-value pairs, construct a
```
SeungheonOh, line 168: It'd be nice to mention here a duplicate entry will get disposed. I think `pinsert` replace the old one with the new one, so it has quite a different behavior as compared to its unsafe counterpart.

src/Plutarch/Extra/Map.hs
```diff
@@ -1,185 +1,266 @@
...
+    (PIsData k, PIsData v) =>
     Term s (k :--> PMap keys k v :--> v)
-plookup' = phoistAcyclic $
-    plam $ \x m -> unTermCont $ do
-        res <- pmatchC (plookup # x # m)
-        pure $ case res of
-            PNothing -> ptraceError "plookup': Could not find key."
+plookupPartial = phoistAcyclic $
+    plam $ \k kvs ->
+        pmatch (plookup # k # kvs) $ \case
```
SeungheonOh, line 139: we have `passertPJust` on `Plutarch.Extra.Maybe`. I'm perfectly fine with either way, but just letting you know there is a shortcut :D

SeungheonOh (changes requested):

SeungheonOh:
+1 for changing the name of `plookupPartial` I was gonna ask to change its name to `ptryLookup`. We had a telepathy :) Besides, few small things.

src/Plutarch/Extra/Map.hs
```diff
@@ -1,185 +1,266 @@
...
+                :--> PBuiltinList (PBuiltinPair (PAsData k) (PAsData b))
+                :--> PBool
+            )
     go = phoistAcyclic $
-        plam $ \x p -> unTermCont $ do
-            k' <- pletC (pfromData $ pfstBuiltin # p)
-            pure $ k' #== x
+        pfix #$ plam $ \self ell ell' ->
+            pmatch (puncons # ell) $ \case
+                PNothing -> pmatch (puncons # ell') $ \case
```
SeungheonOh, line 78: I really think `pfromMaybe` on `Plutarch.Extra.Maybe` would make this code more terse. Personal opinion.

SeungheonOh: ehh. maybe not because it's *very* nested.

src/Plutarch/Extra/Map.hs
SeungheonOh: Oh and also, I think we can use `pelimList` here.

src/Plutarch/Extra/Map.hs
```diff
@@ -1,185 +1,266 @@
...
+                    PNothing -> pcon PTrue -- no mismatches found
+                    PJust _ -> pcon PFalse -- one argument too long
+                PJust kv -> pmatch (puncons # ell') $ \case
+                    PNothing -> pcon PFalse -- one argument too long
+                    PJust kv' -> pmatch kv $ \case
+                        PPair h t -> pmatch kv' $ \case
```
SeungheonOh, line 84: No need for lambda case

src/Plutarch/Extra/Map.hs
```diff
@@ -136,9 +136,8 @@ ptryLookup ::
     Term s (k :--> PMap keys k v :--> v)
 ptryLookup = phoistAcyclic $
     plam $ \k kvs ->
-        pmatch (plookup # k # kvs) $ \case
-            PNothing -> ptraceError "plookupPartial: No value found for key."
-            PJust v -> v
+        passertPJust # "plookupPartial: No value found for key."
+            # (plookup # k # kvs)
```
SeungheonOh, line 140: use `#$`

src/Plutarch/Extra/Map.hs
```diff
@@ -207,8 +211,8 @@ pkeys ::
     (PListLike ell, PElemConstraint ell (PAsData k)) =>
     Term s (PMap keys k v :--> ell (PAsData k))
 pkeys = phoistAcyclic $
-    plam $ \kvs -> pmatch kvs $ \case
```
SeungheonOh: `pmatch` can be used in view patterns like this:
```hs
plam $ \(pmatch -> PMap kvs) -> ...
```

src/Plutarch/Extra/Map.hs
```diff
@@ -80,13 +81,12 @@ pkeysEqual = phoistAcyclic $
...
                 PJust kv -> pmatch (puncons # ell') $ \case
                     PNothing -> pcon PFalse -- one argument too long
-                    PJust kv' -> pmatch kv $ \case
-                        PPair h t -> pmatch kv' $ \case
-                            PPair h' t' ->
-                                pif
-                                    ((pkvPairKey # h) #== (pkvPairKey # h'))
-                                    (self # t # t') -- continue
-                                    (pcon PFalse) -- key mismatch
+                    PJust kv' -> pmatch kv $ \(PPair h t) ->
```
SeungheonOh, line 84: can use viewpattern

SeungheonOh (commented):
Some minor details.

src/Plutarch/Extra/Map.hs
```diff
@@ -1,185 +1,363 @@
...
+padjust ::
+    forall (k :: S -> Type) (v :: S -> Type) (s :: S).
+    (PIsData k, PEq k, PIsData v) =>
+    Term s ((v :--> v) :--> k :--> PMap 'Unsorted k v :--> PMap 'Unsorted k v)
+padjust = phoistAcyclic $
+    plam $ \f key kvs ->
+        pmatch kvs $ \(PMap kvs') ->
+            pcon . PMap $ PList.pmap # (go # f # key) # kvs'
   where
     go ::
```
SeungheonOh, line 71: Having Haskell function here would be cleaner/No need for `phoistAcyclic`. 

```hs
go :: Term s k -> Term s (PBuiltinPair (PAsData k) (PAsData k)) -> PBool
```

SeungheonOh (approved):
There are some `go`s that uses Plutarch function, and some uses Haskell function. I think most can be converted to Haskell function, which I think would have some performance benefits because that doesn't require hoisting. 

Otherwise, LGTM

src/Plutarch/Extra/Map.hs
kozross: Yeah, I agree there, especially since it'll be inlined anyway. Will fix.

PR #73, opened 2022-08-23
Add `Plutarch.Extra.Time`

chfanghr:
This pr is based on #70.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,8 +44,16 @@ otherwise.
 
 ### Added
 
-* A new `Plutarch.Extra.Time` module, containing utilities for working with bounded
-  time range.
+* A `Plutarch.Extra.Time` module, containing utilities for working with time and 
+   time ranges.
+* Some utilities for working with closed bounded time ranges, including
+
+  - `PCurrentTime`
+  - `pcurrentTime`
+  - `currentTime`
+  - `passertCurrentTime`
+  - `pisWithinCurrentTime`
+  - `pisCurrentTimeWithin`
 
 ## 3.2.0 -- 2022-08-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,13 @@ otherwise.
 * `pkeysEqual` now only works for sorted `PMap`s.
 * `pkeysEqual` now exported from `Plutarch.Extra.Map`.
 
+## 3.3.0 -- 2022-08-23
+
+### Added
+
+* A new `Plutarch.Extra.Time` module, containing utilities for working with bounded
+  time range.
+
 ## 3.2.0 -- 2022-08-22
 
 ### Changed
--- a/src/Plutarch/Extra/Time.hs
+++ b/src/Plutarch/Extra/Time.hs
@@ -1,0 +1,159 @@
+{-# LANGUAGE OverloadedRecordDot #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE ViewPatterns #-}
+
+module Plutarch.Extra.Time (
+    PCurrentTime (..),
+    pcurrentTime,
+    currentTime,
+    passertCurrentTime,
+    pisWithinCurrentTime,
+    pisCurrentTimeWithin,
+) where
+
+import Control.Composition ((.*))
+import GHC.Records (HasField)
+import Plutarch.Api.V1 (
+    PExtended (PFinite),
+    PInterval (PInterval),
+    PLowerBound (PLowerBound),
+    PPOSIXTime,
+    PUpperBound (PUpperBound),
+ )
+import Plutarch.Api.V2 (PPOSIXTimeRange)
+import Plutarch.Extra.Applicative (pliftA2)
+import Plutarch.Extra.Field (pletAll, pletAllC)
+import Plutarch.Extra.Maybe (passertPJust, pjust, pnothing)
+import Plutarch.Extra.TermCont (pmatchC)
+
+{- | Represent a fully bounded time range.
+
+     Note: 'PCurrentTime' doesn't need a Haskell-level equivalent because it
+     is only used in scripts, and does not go in datums. It is also scott-encoded
+     which is more efficient in usage.
+
+     @since 3.3.0
+-}
+data PCurrentTime (s :: S) = PCurrentTime
+    { lowerBound :: Term s PPOSIXTime
+    , upperBound :: Term s PPOSIXTime
+    }
+    deriving stock
+        ( -- | @since 3.3.0
+          Generic
+        )
+    deriving anyclass
+        ( -- | @since 3.3.0
+          PlutusType
+        , -- | @since 3.3.0
+          PEq
+        )
+
+-- | @since 3.3.0
+instance DerivePlutusType PCurrentTime where
+    type DPTStrat _ = PlutusTypeScott
+
+{- | Get the current time, given a 'PPOSIXTimeRange'.
+
+     If it's impossible to get a fully-bounded time, (e.g. either end of the 'PPOSIXTimeRange' is
+     an infinity) then we return nothing.
+
+     @since 3.3.0
+-}
+pcurrentTime ::
+    forall (s :: S).
+    Term
+        s
+        ( PPOSIXTimeRange
+            :--> PMaybe PCurrentTime
+        )
+pcurrentTime = phoistAcyclic $
+    plam $ \iv -> unTermCont $ do
+        PInterval iv' <- pmatchC iv
+        ivf <- pletAllC iv'
+        PLowerBound lb <- pmatchC ivf.from
+        PUpperBound ub <- pmatchC ivf.to
+
+        let getBound = phoistAcyclic $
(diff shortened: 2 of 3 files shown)


src/Plutarch/Extra/Time.hs
```diff
@@ -0,0 +1,159 @@
...
+-}
+pisWithinCurrentTime ::
+    forall (s :: S).
+    Term
+        s
+        ( PPOSIXTime
+            :--> PCurrentTime
+            :--> PBool
+        )
+pisWithinCurrentTime = phoistAcyclic $
```
SeungheonOh, line 138: It can be `plam $ \t -> pisCurrentTimeWithin # t` I believe.

src/Plutarch/Extra/Time.hs
```diff
@@ -0,0 +1,159 @@
...
+ )
+import Plutarch.Api.V2 (PPOSIXTimeRange)
+import Plutarch.Extra.Applicative (pliftA2)
+import Plutarch.Extra.Field (pletAll, pletAllC)
+import Plutarch.Extra.Maybe (passertPJust, pjust, pnothing)
+import Plutarch.Extra.TermCont (pmatchC)
+
+{- | Represent a fully bounded time range.
+
+     Note: 'PCurrentTime' doesn't need a Haskell-level equivalent because it
```
SeungheonOh, line 31: Need more explanation the difference between `PCurrentTime` and `PPOSIXTimeRange` because they look identical and have same purpose. More emphasis DataPlut vs ScottPlut would make it more clear.

src/Plutarch/Extra/Time.hs
```diff
@@ -0,0 +1,159 @@
...
+    forall (s :: S).
+    Term
+        s
+        ( PString
+            :--> PPOSIXTimeRange
+            :--> PCurrentTime
+        )
+passertCurrentTime = phoistAcyclic $
+    plam $
+        \msg iv -> passertPJust # msg #$ pcurrentTime # iv
```
SeungheonOh, line 124: eta reduction? I think..

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 This format is based on [Keep A Changelog](https://keepachangelog.com/en/1.0.0).
 
+## 3.3.0 -- 2022-08-23
+
+### Added
+
+* A new `Plutarch.Extra.Time` module, containing utilities for working with bounded
```
SeungheonOh, line 9: New functions should be listed out.

src/Plutarch/Extra/Time.hs
```diff
@@ -0,0 +1,159 @@
...
+{- | Represent a fully bounded time range.
+
+     Note: 'PCurrentTime' doesn't need a Haskell-level equivalent because it
+     is only used in scripts, and does not go in datums. It is also scott-encoded
+     which is more efficient in usage.
+
+     @since 3.3.0
+-}
+data PCurrentTime (s :: S) = PCurrentTime
+    { lowerBound :: Term s PPOSIXTime
```
SeungheonOh, line 38: I feel like we really don't need a record here since it's gonna be `pmatch`ed and users will pattern match it anyways. Happy to keep it here if there is a good reason I'm unaware of.

SeungheonOh (changes requested):
It's a bit late. Sorry. A few changes and questions but otherwise looks good!

src/Plutarch/Extra/Time.hs
kozross: My guess is record labels.

src/Plutarch/Extra/Time.hs
chfanghr: Note that `PPOSIXTimeRange` doesn't require a time range to be bounded and closed on both sides. And also as you pointed it out, `PCurrentTime` is scott-encoded.

src/Plutarch/Extra/Time.hs
chfanghr: I don't think it can. `pisWithinCurrentTime` tell you a point of time is in the current time range, while  `pisCurrentTimeWithin` is supposed to yield true if current time range is in the given time range - the first argument of `pisCurrentTimeWithin` is actually the lower bound of the said time range.

src/Plutarch/Extra/Time.hs
chfanghr: Kinda wish that we have plutarch level dot.

src/Plutarch/Extra/Time.hs
SeungheonOh: Yeah. I'd be nice to mention that in docstring.

src/Plutarch/Extra/Time.hs
SeungheonOh: Okay. It's way more confusing to eta reduce this. Well, I guess I shouldn't call it a reduction in the first place lol

src/Plutarch/Extra/Time.hs
```diff
@@ -0,0 +1,164 @@
...
+import Plutarch.Api.V2 (PPOSIXTimeRange)
+import Plutarch.Extra.Applicative (pliftA2)
+import Plutarch.Extra.Field (pletAll, pletAllC)
+import Plutarch.Extra.Maybe (passertPJust, pjust, pnothing)
+import Plutarch.Extra.TermCont (pmatchC)
+
+{- | Represent a fully bounded time range.
+
+     Note: 'PCurrentTime' doesn't need a Haskell-level equivalent because it
+     is only used in scripts, and does not go in datums. It is also scott-encoded
```
kozross, line 32: 'Scott-encoded'

src/Plutarch/Extra/Time.hs
```diff
@@ -0,0 +1,164 @@
...
+     which is more efficient in usage.
```
kozross, line 33: I'd just remove this line.

src/Plutarch/Extra/Time.hs
```diff
@@ -0,0 +1,164 @@
...
+        )
+
+-- | @since 3.3.0
+instance DerivePlutusType PCurrentTime where
+    type DPTStrat _ = PlutusTypeScott
+
+{- | Get the current time, given a 'PPOSIXTimeRange'.
+
+     If it's impossible to get a fully-bounded time, (e.g. either end of the 'PPOSIXTimeRange' is
+     an infinity) then we return nothing.
```
kozross, line 61: "If it's impossible to get a full-bounded time (for example, either end of the 'PPOSIXTimeRange' is an infinity), then we return 'PNothing'."

src/Plutarch/Extra/Time.hs
```diff
@@ -0,0 +1,164 @@
...
+                            )
+                            (ptrace "pcurrentTime: time range should be inclusive" pnothing)
+
+            lb' = getBound # lb
+            ub' = getBound # ub
+
+            mkTime = phoistAcyclic $ plam $ pcon .* PCurrentTime
+        pure $ pliftA2 # mkTime # lb' # ub'
+
+{- | Calculate current time providing the @validaRange@ field, which typically
```
kozross, line 99: "Calculate the current time by providing the @validRange@ field, ..."

src/Plutarch/Extra/Time.hs
```diff
@@ -0,0 +1,164 @@
...
+-}
+currentTime ::
+    forall r (s :: S).
+    (HasField "validRange" r (Term s PPOSIXTimeRange)) =>
+    r ->
+    Term s (PMaybe PCurrentTime)
+currentTime x = pcurrentTime # x.validRange
+
+{- | Calculate current time, and error out with the given message if we can't
+      get one.
```
kozross, line 112: "the current time"

"if we can't get it"

src/Plutarch/Extra/Time.hs
```diff
@@ -0,0 +1,164 @@
...
+        s
+        ( PString
+            :--> PPOSIXTimeRange
+            :--> PCurrentTime
+        )
+passertCurrentTime = phoistAcyclic $
+    plam $
+        \msg iv -> passertPJust # msg #$ pcurrentTime # iv
+
+{- | Retutn true if a `PPOSIXTime` is in the current time range.
```
kozross, line 128: "Return 'PTrue' if ..."

src/Plutarch/Extra/Time.hs
```diff
@@ -0,0 +1,164 @@
...
+        )
+pisWithinCurrentTime = phoistAcyclic $
+    plam $ \time ctr ->
+        pmatch ctr $ \(PCurrentTime lb ub) ->
+            lb #<= time #&& time #<= ub
+
+{- | Return true if current time is within the given time range.
+
+    Note that the first argument is the lower bound of the said time range, and
+      the second is the upper bound.
```
kozross, line 148: "Return 'PTrue' if the current time .."

"of said time range..."

kozross (changes requested):
Some grammar issues in the docs.

kozross (approved):

SeungheonOh (approved):
LGTM.

PR #76, opened 2022-08-28
Remove sorting-related functionality from Extra.List

kozross:
This removes all sort-related functionality from `Plutarch.Extra.List`, as well as cleaning up and generalizing some other functions. We also no longer re-export from `plutarch-extra`'s version of the same module, in line with future upstreaming efforts.


src/Plutarch/Extra/List.hs
```diff
@@ -1,380 +1,151 @@
...
-                let y = phead # xs
-                 in plet (ptail # xs) $ \ys ->
-                        pif (pnull # ys) xs $
-                            let z = phead # ys
-                                zs = ptail # ys
-                             in pcons # (pmergeBy # comp # y # z) # (self # comp # zs)
+    forall (a :: S -> Type) (ell :: (S -> Type) -> S -> Type) (s :: S).
+    (PIsListLike ell a) =>
+    Term s (ell a :--> PBool)
+pnotNull = phoistAcyclic $ plam $ \xs -> pnot #$ pnull # xs
```
SeungheonOh, line 38: Do we really need this function?

src/Plutarch/Extra/List.hs
```diff
@@ -1,380 +1,151 @@
...
+    forall (b :: S -> Type) (ell :: (S -> Type) -> S -> Type) (a :: S -> Type) (s :: S).
+    (PElemConstraint ell a, PElemConstraint ell b, PListLike ell) =>
+    Term s ((a :--> PMaybe b) :--> ell a :--> ell b)
 pmapMaybe = phoistAcyclic $
-    pfix #$ plam $ \self f l -> pif (pnull # l) pnil $
-        unTermCont $ do
-            x <- pletC $ phead # l
-            xs <- pletC $ ptail # l
+    pfix #$ plam $ \self f xs ->
+        pmatch (PList.puncons # xs) $ \case
```
SeungheonOh, line 53: `pelimList` might have a better performance here.

src/Plutarch/Extra/List.hs
```diff
@@ -1,380 +1,151 @@
...
-                    PJust ux -> pcons # ux #$ self # f # xs
-                    _ -> self # f # xs
+{- | As 'pfind', but with a Haskell-level function on 'Term's for the predicate.
 
-{- | Get the first element that matches a predicate or return Nothing.
-
-     @since 1.1.0
+ @since 3.4.0
 -}
 pfind' ::
```
SeungheonOh, line 64: We can really drop this and use `pfind`.

src/Plutarch/Extra/List.hs
```diff
@@ -1,380 +1,151 @@
...
             pif
-                (pnull # xs)
-                (pconstant True)
-                $ plet (phead # xs) $ \x' ->
-                    pif
-                        (eq # x # x' #|| lt # x # x')
-                        (self # x' #$ ptail # xs)
-                        (pconstant False)
+                (target #== (getKey # kv))
+                (pcon . PJust $ kv)
```
SeungheonOh, line 128: neat pick:
`pcon $ PJust kv`

src/Plutarch/Extra/Ord.hs
```diff
@@ -0,0 +1,280 @@
...
+    stimes = stimesIdempotentMonoid
+
+-- | @since 3.4.0
+instance Monoid (Term s POrdering) where
+    mempty = pcon PEQ
+
+-- TODO: PShow, PPartialOrd, POrd
+
+-- | @since 3.4.0
+newtype PComparator (a :: S -> Type) (s :: S)
```
SeungheonOh, line 136: Ideally we have to change `POrd`...

src/Plutarch/Extra/Ord.hs
```diff
@@ -0,0 +1,280 @@
...
+
+ This assumes that \'@c@s that are @a@s\' should be ordered before \'@c@s that
+ are @b@s\'.
+
+ @since 3.4.0
+-}
+psumComparator ::
+    forall (a :: S -> Type) (b :: S -> Type) (c :: S -> Type) (s :: S).
+    Term s ((c :--> PEither a b) :--> PComparator a :--> PComparator b :--> PComparator c)
+psumComparator = phoistAcyclic $
```
SeungheonOh, line 217: `TermCont` might make a bit more cleaner code. But it's good as is.

SeungheonOh: Actually, since I see a lot of `pmatch comp $ \(PComparator x) -> ...` pattern, it might be a good idea to make a function for this. 

Or, `pto` works as well.

src/Plutarch/Extra/Ord.hs
```diff
@@ -0,0 +1,280 @@
...
+ instance to construct the 'PComparator'. Allows other \'-by\' behaviours.
+
+ @since 3.4.0
+-}
+pfromOrdBy ::
+    forall (a :: S -> Type) (b :: S -> Type) (s :: S).
+    (POrd a) =>
+    Term s ((b :--> a) :--> PComparator b)
+pfromOrdBy = phoistAcyclic $
+    plam $ \f -> pcon . PComparator $
```
SeungheonOh, line 175: `TermCont`

SeungheonOh (changes requested):

src/Plutarch/Extra/List.hs
kozross: I left it in for compatibility. I don't see the harm in having it, but I agree, it's a bit superfluous.

src/Plutarch/Extra/List.hs
kozross: Yeah, I'm not too sure why it's here either.

src/Plutarch/Extra/Ord.hs
kozross: Ideally, but that can wait until we upstream.

src/Plutarch/Extra/Ord.hs
kozross: `pto` is probably what I was looking for, thanks!

SeungheonOh: `pto` is so helpful but I always forget it's there when I need it...

src/Plutarch/Extra/Boost.hs
```diff
@@ -0,0 +1,58 @@
...
+    -- * CPS-driven matching
+    pmatchList,
+) where
+
+{- | Make a list-like structure from any 'Foldable' of 'Term's. This is
+ unhoisted.
+
+ @since 3.4.0
+-}
+plist ::
```
SeungheonOh, line 25: These belong on `Plutarch.Extra.List` I think. Or, if it's only used by the sort, they should go into `Ord` I think.

SeungheonOh (approved):
There are things that I don't understand entirely, like the sorting network part and some merging internals, but otherwise looks great.

SeungheonOh:
Do we want another review from @cjay?

kozross:
It can't hurt - added.

src/Plutarch/Extra/Ord.hs
```diff
@@ -0,0 +1,641 @@
...
+        precList (go cmp) (const (pcon . PJust . pconstant $ True))
+  where
+    go ::
+        forall (s' :: S).
+        Term s' (PComparator a) ->
+        Term s' (ell a :--> PMaybe PBool) ->
+        Term s' a ->
+        Term s' (ell a) ->
+        Term s' (PMaybe PBool)
+    go cmp self x xs = pmatch (PList.puncons # xs) $ \case
```
cjay, line 214: This looks very inefficient. Not only does it peek ahead instead of remembering using an extra parameter of the recursive call, it also creates a `PMaybe` value for each element and matches on that.

cjay (changes requested):
Lots of problems. i'd also get rid of `puncons` everywhere.

src/Plutarch/Extra/Ord.hs
```diff
@@ -0,0 +1,641 @@
...
+    Term s (ell a)
+psingletonUnhoisted x = plist [x]
+
+-- Turns any Foldable into a list-like, unhoisted
+plist ::
+    forall (ell :: (S -> Type) -> S -> Type) (f :: Type -> Type) (a :: S -> Type) (s :: S).
+    (Foldable f, PListLike ell, PElemConstraint ell a) =>
+    f (Term s a) ->
+    Term s (ell a)
+plist = foldr (\x xs -> pcons # x # xs) pnil
```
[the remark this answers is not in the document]
kozross: Is there a way I can _ensure_ un-hoistedness?

src/Plutarch/Extra/Ord.hs
```diff
@@ -0,0 +1,641 @@
...
+                \_1 _2 -> plist [_0, _1, _2]
+    psort4 ::
+        forall (s' :: S).
+        Term s' (PComparator a) ->
+        Term s' a ->
+        Term s' a ->
+        Term s' a ->
+        Term s' a ->
+        Term s' (ell a)
+    psort4 cmp _0 _1 _2 _3 = pswap cmp _0 _2 $
```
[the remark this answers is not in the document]
kozross: I'd prefer not to use it, because I don't think it's _that_ unreadable.

src/Plutarch/Extra/Ord.hs
```diff
@@ -0,0 +1,641 @@
...
+                PNothing -> xs
+                PJust ys' -> pmatch xs' $ \(PPair leftH leftT) ->
+                    pmatch ys' $ \(PPair rightH rightT) ->
+                        pmatch (pcompareBy # cmp # leftH # rightH) $ \case
+                            -- Right before left.
+                            PGT -> pcons # rightH #$ pcons # leftH # (self # cmp # leftT # rightT)
+                            -- Left before right.
+                            _ -> pcons # leftH #$ pcons # rightH # (self # cmp # leftT # rightT)
+
+pmergeStart_2_3 ::
```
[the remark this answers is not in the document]
kozross: That's fine. It should be named something more meaningful anyway.

src/Plutarch/Extra/Ord.hs
```diff
@@ -0,0 +1,641 @@
...
+pmergeAll = phoistAcyclic $
+    pfix #$ plam $ \self cmp xss ->
+        pmatch (PList.puncons # xss) $ \case
+            PNothing -> pnil
+            PJust xss' -> pmatch xss' $ \(PPair h t) ->
+                pmatch (PList.puncons # t) $ \case
+                    PNothing -> h
+                    PJust _ -> self # cmp #$ go # cmp # t
+  where
+    go ::
```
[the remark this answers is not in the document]
kozross: `go` is the generic name for a workhorse function in a `where`-bind by Haskell convention.

src/Plutarch/Extra/Ord.hs
```diff
@@ -0,0 +1,641 @@
...
+        forall (s' :: S).
+        Term s' (PComparator a :--> ell (ell a) :--> ell (ell a))
+    go = phoistAcyclic $
+        pfix #$ plam $ \self cmp xss ->
+            pmatch (PList.puncons # xss) $ \case
```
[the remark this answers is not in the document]
kozross: This is a good point: `pelimList` is probably a better choice here.

src/Plutarch/Extra/Ord.hs
```diff
@@ -0,0 +1,641 @@
...
+    (PElemConstraint ell a, PElemConstraint ell (ell a), PListLike ell) =>
+    Term s (PComparator a :--> ell (ell a) :--> ell a)
+pmergeAll = phoistAcyclic $
+    pfix #$ plam $ \self cmp xss ->
+        pmatch (PList.puncons # xss) $ \case
+            PNothing -> pnil
+            PJust xss' -> pmatch xss' $ \(PPair h t) ->
+                pmatch (PList.puncons # t) $ \case
+                    PNothing -> h
+                    PJust _ -> self # cmp #$ go # cmp # t
```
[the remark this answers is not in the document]
kozross: This all needs tests, to be fair. Nice catch!

src/Plutarch/Extra/Ord.hs
```diff
@@ -0,0 +1,641 @@
...
+{- | As 'pmerge', but specifying the ordering by a custom 'PComparator'.
+
+ @since 3.5.0
+-}
+ptryMergeBy ::
+    forall (a :: S -> Type) (ell :: (S -> Type) -> S -> Type) (s :: S).
+    (PElemConstraint ell a, PListLike ell) =>
+    Term s (PComparator a :--> ell a :--> ell a :--> ell a)
+ptryMergeBy = phoistAcyclic $
+    pfix #$ plam $ \self cmp xs ys ->
```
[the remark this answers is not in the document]
kozross: I get some issues with unification here, as they have different `S`es.

src/Plutarch/Extra/Ord.hs
```diff
@@ -0,0 +1,641 @@
...
+                PEQ -> xs
+                -- Keep.
+                _ -> pcons # x # xs
+
+pmergeAll ::
+    forall (a :: S -> Type) (ell :: (S -> Type) -> S -> Type) (s :: S).
+    (PElemConstraint ell a, PElemConstraint ell (ell a), PListLike ell) =>
+    Term s (PComparator a :--> ell (ell a) :--> ell a)
+pmergeAll = phoistAcyclic $
+    pfix #$ plam $ \self cmp xss ->
```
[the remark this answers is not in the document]
kozross: Same here.

src/Plutarch/Extra/Ord.hs
```diff
@@ -0,0 +1,641 @@
...
+                            pcons # (pmergeUnsafe # cmp # h # h') # (self # cmp # t')
+
+-- This assumes both argument list-likes are sorted by the argument comparator:
+-- be careful how you call this!
+pmergeUnsafe ::
+    forall (a :: S -> Type) (ell :: (S -> Type) -> S -> Type) (s :: S).
+    (PElemConstraint ell a, PListLike ell) =>
+    Term s (PComparator a :--> ell a :--> ell a :--> ell a)
+pmergeUnsafe = phoistAcyclic $
+    pfix #$ plam $ \self cmp xs ys ->
```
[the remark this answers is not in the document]
kozross: And here.

src/Plutarch/Extra/Ord.hs
```diff
@@ -0,0 +1,641 @@
...
+
+ @since 3.5.0
+-}
+ptryMerge ::
+    forall (a :: S -> Type) (ell :: (S -> Type) -> S -> Type) (s :: S).
+    (PElemConstraint ell a, PListLike ell, POrd a) =>
+    Term s (ell a :--> ell a :--> ell a)
+ptryMerge = phoistAcyclic $ plam $ \xs -> ptryMergeBy # pfromOrd @a # xs
+
+{- | As 'pmerge', but specifying the ordering by a custom 'PComparator'.
```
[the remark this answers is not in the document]
kozross: I guess I could clear this up? I figured referencing `pmerge` would be sufficient, but seems I was wrong.

src/Plutarch/Extra/Ord.hs
```diff
@@ -0,0 +1,804 @@
...
+pmergeAllUnsafe ::
+    forall (a :: S -> Type) (ell :: (S -> Type) -> S -> Type) (s :: S).
+    (PElemConstraint ell a, PListLike ell) =>
+    Term s (PComparator a) ->
+    (forall (s' :: S). Term s' (PComparator a) -> Term s' (PList a) -> Term s' (ell a)) ->
+    (forall (s' :: S). Term s' (PComparator a) -> Term s' (PList a) -> Term s' (PList a) -> Term s' (PList a)) ->
+    Term s (PList (PList a) :--> ell a)
+pmergeAllUnsafe cmp whenSingleton whenMerging = pfix #$ plam $ \self xs ->
+    phandleList xs pnil $ \x' xs' ->
+        phandleList xs' (whenSingleton cmp x') $ \_ _ ->
```
cjay, line 560: `whenSingleton` is pointless, same efficiency as doing `pconvertLists` after the fact. Should produce the output list type when doing the last merge instead. You could probably even do this with the same `whenMerging` if it's polymorphic in its output list type, it just needs to be independent of its input list type.

src/Plutarch/Extra/Ord.hs
```diff
@@ -0,0 +1,804 @@
...
+pmergeUnsafe cmp xs ys =
+    phandleList xs ys $ \x' xs' ->
+        phandleList ys xs $ \y' ys' ->
+            (pfix #$ plam $ go) # x' # xs' # y' # ys'
+  where
+    go ::
+        Term s (a :--> PList a :--> a :--> PList a :--> PList a) ->
+        Term s a ->
+        Term s (PList a) ->
+        Term s a ->
```
cjay, line 588: This can be done with one less argument, see my `pmerge`.

src/Plutarch/Extra/Ord.hs
```diff
@@ -0,0 +1,804 @@
...
+pmergeUnsafeNoDupes cmp xs ys =
+    phandleList xs (pnubUnsafe # cmp # ys) $ \x' xs' ->
+        phandleList ys (pnubUnsafe # cmp # xs) $ \y' ys' ->
+            (pfix #$ plam $ go) # x' # xs' # y' # ys'
+  where
+    go ::
+        Term s (a :--> PList a :--> a :--> PList a :--> PList a) ->
+        Term s a ->
+        Term s (PList a) ->
+        Term s a ->
```
cjay, line 617: Same here.

src/Plutarch/Extra/Ord.hs
```diff
@@ -0,0 +1,804 @@
...
+                    plam $
+                        \cmp' _0 _1 _2 _3 ->
+                            pswap cmp' _0 _1 $ \_0 _1 ->
+                                pswap cmp' _2 _3 $ \_2 _3 ->
+                                    pswap cmp' _1 _2 $ \_1 _2 ->
+                                        plist [_0, _1, _2, _3]
+
+-- Runs a 'sorting network layer' driven by a given PComparator. Written in CPS
+-- style for efficiency.
+pswap ::
```
cjay, line 755: Should include most of my docstring.

src/Plutarch/Extra/Ord.hs
```diff
@@ -0,0 +1,804 @@
...
+        Term s' (PList a)
+    go cmp x peek rest =
+        pif
+            (pequateBy # cmp # x # peek)
+            rest
+            (pcons # x # rest)
+
+-- Breaks the argument into sorted chunks; currently, these go up to size 4. To
+-- speed up the chunk sorting, we use sorting networks.
+pmergesort ::
```
cjay, line 654: Name is very misleading, its just the starting phase, and a very specific one (with networks up to size 4). Better leave it open to add different alternative starting phases. Should be named `pmergeStart_2_3_4`.

cjay (changes requested):
One bug and still a few inefficient things.

src/Plutarch/Extra/Ord.hs
cjay: Someone with input sizes of only 64 and 12 doesn't need the smaller networks, for example. On the other hand, someone with input size 40 might want network 5, to get 8 lists that can be merged without leftovers (power of 2). So different starting phases for different use cases make sense.

cjay:
Strange, github repeated some of my comments from the previous review.

src/Plutarch/Extra/Ord.hs
```diff
@@ -0,0 +1,804 @@
...
+    forall (a :: S -> Type) (r :: S -> Type) (s :: S).
+    Term s (PComparator a) ->
+    Term s a ->
+    Term s a ->
+    (Term s a -> Term s a -> Term s r) ->
+    Term s r
+pswap cmp x y cont = pif (pleqBy # cmp # x # y) (cont x y) (cont y x)
+
+-- Similar to zipWith f xs (tail xs)
+precList2 ::
```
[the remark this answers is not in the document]
kozross: I agree, the name is misleading. What would you suggest instead?

src/Plutarch/Extra/Ord.hs
```diff
@@ -0,0 +1,804 @@
...
+        forall (s' :: S).
+        Term s' (PComparator a) ->
+        Term s' (a :--> ell a :--> a :--> ell a :--> ell a) ->
+        Term s' a ->
+        Term s' (ell a) ->
+        Term s' a ->
+        Term s' (ell a) ->
+        Term s' (ell a)
+    go cmp self x xs y ys = pmatch (pcompareBy # cmp # x # y) $ \case
+        PGT -> phandleList ys (pcons # y #$ passertSorted # cmp # xs) $ \y' ys' ->
```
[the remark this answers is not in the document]
kozross: Thanks! That's definitely getting pinned in place today.

src/Plutarch/Extra/Ord.hs
kozross: Yeah, fair enough.

kozross:
@cjay Github is _super_ weird when it comes to comments on PRs. I've had it do all sorts of things, including breaking links.

src/Plutarch/Extra/Ord.hs
```diff
@@ -0,0 +1,804 @@
...
+        PEQ -> pmatch y $ \case
+            PEQ -> pcon PTrue
+            _ -> pcon PFalse
+        PGT -> pmatch y $ \case
+            PGT -> pcon PTrue
+            _ -> pcon PFalse
+
+-- | @since 3.6.0
+instance Semigroup (Term s POrdering) where
+    x <> y = pmatch x $ \case
```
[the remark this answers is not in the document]
kozross: Given that we need benches for this stuff in general, it might be a good idea to bench this to be sure.

bench/sorting/Main.hs
```diff
@@ -0,0 +1,95 @@
...
+    singleTest
+        ("pnubSort, custom comparator, length " <> show len)
+        ( ScriptBench $
+            pnubSortBy @PInteger @PList # (preverseComparator # pfromOrd)
+                #$ preplicate # plen # 1
+        )
+
+-- Scaffolding and helpers
+
+newtype ScriptBench (a :: S -> Type) = ScriptBench (forall (s :: S). Term s a)
```
SeungheonOh, line 71: We are having this _**same**_ newtype wrappers all over the place. :rofl: It'd be nice to have newtyped `ClosedTerm`...

bench/sorting/Main.hs
```diff
@@ -49,38 +49,44 @@ benchNub :: (forall (s :: S). Term s PInteger) -> Int -> TestTree
 benchNub plen len =
     singleTest
         ("pnubSort, length " <> show len)
-        (ScriptBench $ pnubSort @PInteger @PList #$ preplicate # plen # 1)
+        (ScriptBench (pnubSort @PInteger @PList, preplicate # plen # 1))
```
SeungheonOh, line 52: type applications with tick mark should be put into a parenthesis because of fourmolu bug.. 

like this
```hs
@( 'PList)
```

I'm not sure if we are enforcing ticked mark on LPE..

kozross: I don't think type applications normally _get_ ticks. At the very least, I haven't been using them.

cjay (changes requested):
Looks good, except for `whenSingleton` in `pmergeAllUnsafe`. The last merge could create the desired output list type directly.

PR #77, opened 2022-08-31
Fix issue 60

### LINKED ISSUE #60: `pfindDatum` is broken
The type of the datum map changed from V1 to V2. I'm going to write a local function in xplorerdao to make sure things work and then upstream it here.

kozross:
This is meant to close #60. Upon discussion with @peter-mlabs, it seems we already have a function with the desired capabilities in `presolveOutputDatum`. This PR adds an 'erroring' counterpart `ptryResolveOutputDatum`, and removes `pfindDatum` and its 'erroring' counterpart.


peter-mlabs (approved):
Looks good to me.

PR #80, opened 2022-09-01
`PlutusTypeDataList` ignores `PAsData`

SeungheonOh:
This will make `PlutusTypeDataList` not complain about the difference between, for example, `PBuiltinList (PAsData PInteger)` and `PBuiltinList PInteger`.

It's helpful because `PTryFrom` only works with former.


chfanghr (approved):
Work with agora: https://github.com/Liqwid-Labs/agora/commit/96664195646a79f0a68be6ea2bfba4d9b36666cd

PR #84, opened 2022-09-20
Add Aeson instances

peter-mlabs:


kozross (changes requested):

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 This format is based on [Keep A Changelog](https://keepachangelog.com/en/1.0.0).
 
+## 3.6.1 -- 2022-09-20
+
+### Added
+
+* `Aeson` instances for Plutus types (in the `Plutarch.Orphans` module)
```
kozross, line 9: Note the `newtype` being exported from that module, as this is fairly unusual. Normally, `Orphans` modules don't export anything.

src/Plutarch/Orphans.hs
```diff
@@ -62,3 +102,165 @@ instance PTryFrom PData (PAsData PScriptHash) where
...
+        (Aeson.FromJSON MintingPolicy)
+
+deriving via
+    (AsBase16Codec Script)
+    instance
+        (Aeson.ToJSON Script)
+deriving via
+    (AsBase16Codec Script)
+    instance
+        (Aeson.FromJSON Script)
```
kozross, line 266: All require `since`s.

kozross (changes requested):

CHANGELOG.md
peter-mlabs: I don't think we actually need that export...

peter-mlabs: I'll check

src/Plutarch/Orphans.hs
peter-mlabs: ooft. I noticed this, went to add to the change log, and then immediately forgot :sweat_smile:

CHANGELOG.md
peter-mlabs: Don't need it. Its been removed.

kozross (approved):

PR #85, opened 2022-09-20
Extra units

peter-mlabs:


src/Plutarch/Extra/AssetClass.hs
```diff
@@ -1,97 +1,361 @@
...
+
+--------------------------------------------------------------------------------
+-- AssetClass & Variants
+
+{- | Version of Plutus AssetClass, with currency unit tag
+ @since 3.8.0
+-}
+data AssetClass (tag :: Symbol) = AssetClass
+    { symbol :: Plutus.CurrencySymbol
+    , name :: Plutus.TokenName
```
kozross, line 100: Need `since`s on these, since they only appeared under their current names in this version. Also, would `CurrencySymbol` and `TokenName` conflict with something? If so, they should be imported unqualified.

src/Plutarch/Extra/AssetClass.hs
```diff
@@ -1,97 +1,361 @@
...
+    deriving anyclass
+        ( -- | @since 3.8.0
+          ToJSON
+        , -- | @since 3.8.0
+          FromJSON
+        , -- | @since 3.8.0
+          FromJSONKey
+        , -- | @since 3.8.0
+          ToJSONKey
+        )
```
kozross, line 121: Did any of these instances exist on the old `AssetClass`? If so, they shouldn't be tagged as `since` 3.8.0. Only new instance declarations should be.

src/Plutarch/Extra/AssetClass.hs
```diff
@@ -1,97 +1,361 @@
...
+    deriving anyclass
+        ( -- | @since 3.8.0
+          PEq
+        , -- | @since 3.8.0
+          PlutusType
+        )
+
+-- | @since 3.8.0
+instance DerivePlutusType (PAssetClass unit) where
+    type DPTStrat _ = PlutusTypeScott
```
kozross, line 143: Same as above.

src/Plutarch/Extra/AssetClass.hs
```diff
@@ -1,97 +1,361 @@
...
+pcoerceCls = punsafeCoerce
+
+{- | Construct a PAssetClass with empty TokenName
+ | @since 3.8.0
+-}
+psymbolAssetClass ::
+    forall (unit :: Symbol) (s :: S).
+    Term s (PAsData PCurrencySymbol) ->
+    PAssetClass unit s
+psymbolAssetClass sym = PAssetClass sym emptyTokenNameData
```
kozross, line 170: "'PAssetClass'" and "'TokenName'" so the Haddocs get the links.

src/Plutarch/Extra/AssetClass.hs
```diff
@@ -1,97 +1,361 @@
...
+
+--------------------------------------------------------------------------------
+-- Simple Helpers
+
+{- | Check whether an AssetClass is the Ada Class
```
kozross, line 175: Same here. In fact, it's actually quite important to correctly-link Haddocks here, since `AssetClass` exists in `plutus-tx` as well.

src/Plutarch/Extra/AssetClass.hs
```diff
@@ -1,97 +1,361 @@
...
+ | @since 3.8.0
+-}
+adaSymbolData :: forall (s :: S). Term s (PAsData PCurrencySymbol)
+adaSymbolData = pconstantData ""
+
+{- | Ada AssetClass - ("", "")
+ | @since 3.8.0
+-}
+adaClass :: AssetClass "Ada"
+adaClass = AssetClass "" ""
```
kozross, line 193: Not quite sure what these Haddocks are saying.

src/Plutarch/Extra/AssetClass.hs
```diff
@@ -1,97 +1,361 @@
...
+
+----------------------------------------
+-- Data-Encoded version
+
+{- | Data-encoded AssetClass without the phantom-tag
+ | @since 3.8.0
+-}
+data AssetClassData = AssetClassData
+    { symbol :: Plutus.CurrencySymbol
+    , name :: Plutus.TokenName
```
kozross, line 215: These need `since`s too.

src/Plutarch/Extra/AssetClass.hs
```diff
@@ -1,97 +1,361 @@
...
+
+{- | Convert from `AssetClass` to `AssetClassData`
+ | @since 3.8.0
+-}
+fromScottEncoding :: forall (tag :: Symbol). AssetClass tag -> AssetClassData
+fromScottEncoding (AssetClass sym tk) = AssetClassData sym tk
+
+{- | Convert from `PAssetClassData` to `PAssetClass`
+ @since 3.8.0
+-}
```
kozross, line 307: These need 'quote wrapping' so Haddocks link.

src/Plutarch/Extra/AssetClass.hs
```diff
@@ -1,97 +1,361 @@
...
+
+{- | Version of assetClassValue for tagged AssetClass & Tagged
+ @since 3.8.0
+-}
+assetClassValue ::
+    forall (unit :: Symbol).
+    AssetClass unit ->
+    Tagged unit Integer ->
+    Value.Value
+assetClassValue (AssetClass sym tk) q = Value.singleton sym tk $ untag q
```
kozross, line 361: Same here.

src/Plutarch/Extra/ExchangeRate.hs
```diff
@@ -0,0 +1,85 @@
...
+import Plutarch.Extra.Applicative (ppure)
+import Plutarch.Extra.Rational (divRational, divTruncate, mulRational, mulTruncate)
+import Plutarch.Extra.Tagged (PTagged)
+
+{- | Represents an exchange from a to b.
+     Let's say 1.00 ADA is worth 2.00 USD, then @ADA ':>' USD@ ought to be
+     represented as 2.00.
+
+   @since 3.8.0
+-}
```
kozross, line 22: `@a@` and `@b@` so we get monospace.

src/Plutarch/Extra/ExchangeRate.hs
```diff
@@ -0,0 +1,85 @@
...
+exchangeToTruncate ::
+    forall (a :: Symbol) (b :: Symbol) (s :: S).
+    Term
+        s
+        ( PTagged (a :> b) PRational :--> PTagged b PInteger
+            :--> PTagged a PInteger
+        )
+exchangeToTruncate =
+    phoistAcyclic $
+        plam $ \ex x -> ppure #$ divTruncate # pto ex # pto x
```
kozross, line 54: I'm not sure I follow why we need both. They seem to only differ in variable naming in their signatures. Does `:<` imply a sizing relationship?

src/Plutarch/Extra/ExchangeRate.hs
```diff
@@ -0,0 +1,85 @@
...
+
+{- | Convert between quantities of currencies using a Rational conversion value
+
+ @since 3.8.0
+-}
```
kozross, line 59: "'PRational'", right?

src/Plutarch/Extra/Rational.hs
```diff
@@ -0,0 +1,122 @@
+{-# LANGUAGE PolyKinds #-}
+-- TODO: Remove this and replace with TermCont.
+{-# LANGUAGE QualifiedDo #-}
+{-# LANGUAGE RankNTypes #-}
+-- TODO: Either disable warning about orphans altogether or address the issue
+-- requiring this unusual solution.
+{-# OPTIONS_GHC -Wwarn=orphans #-}
```
kozross, line 7: These need addressing.

src/Plutarch/Extra/Rational.hs
```diff
@@ -0,0 +1,122 @@
...
+divTruncate =
+    phoistAcyclic $
+        plam $ \ex x -> P.do
+            (PRational num denom) <- pmatch ex
+            mulDivTruncate # x # pto denom # num
+
+{- | Multiply a Rational by an Integer, without reducing the fraction
+
+ @since 3.8.0
+-}
```
kozross, line 74: This is _definitely_ in need of giant screaming warnings I think, since that can cause surprising perf blowups down the line unexpectedly.

src/Plutarch/Extra/Rational.hs
```diff
@@ -0,0 +1,122 @@
...
+mulRational =
+    phoistAcyclic $
+        plam $ \x r -> P.do
+            (PRational num denom) <- pmatch r
+            pcon $ PRational (num * x) denom
+
+{- | Multiply a Rational by an Integer, without reducing the fraction
+
+ @sinmce 3.8.0
+-}
```
kozross, line 87: `since`

Also, weird argument order.

src/Plutarch/Extra/Rational.hs
```diff
@@ -0,0 +1,122 @@
...
+(#%) ::
+    forall (s :: S).
+    Term s PInteger ->
+    Term s PInteger ->
+    Term s PRational
+x #% y =
+    pif
+        (y #< 0)
+        (pcon $ PRational (x * (-1)) (ptryPositive # (y * (-1))))
+        (pcon $ PRational x (ptryPositive # y))
```
kozross, line 111: Needs fixity declaration. Should probably match the equivalent in `base`.

src/Plutarch/Extra/ScriptContext.hs
```diff
@@ -147,7 +149,7 @@ pvalueSpent = phoistAcyclic $
 
     @since 1.1.0
 -}
-pisTokenSpent :: forall (s :: S). Term s (PAssetClass :--> PBuiltinList PTxInInfo :--> PBool)
+pisTokenSpent :: forall (tag :: Symbol) (s :: S). Term s (PAssetClass tag :--> PBuiltinList PTxInInfo :--> PBool)
```
kozross, line 152: API change, so needs a `since` bump. Also a changelog entry if it doesn't already have one.

src/Plutarch/Extra/Value.hs
```diff
@@ -1,6 +1,12 @@
+{-# LANGUAGE DuplicateRecordFields #-}
+{-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE OverloadedRecordDot #-}
+{-# LANGUAGE PartialTypeSignatures #-}
```
kozross, line 4: Resolve this first.

src/Plutarch/Extra/Value.hs
```diff
@@ -1,6 +1,12 @@
+{-# LANGUAGE DuplicateRecordFields #-}
+{-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE OverloadedRecordDot #-}
```
kozross, line 3: Do we need these extensions? I'm particularly wary of `DuplicateRecordFields` if it can at all be avoided.

src/Plutarch/Extra/Value.hs
```diff
@@ -12,8 +18,28 @@ module Plutarch.Extra.Value (
     pgeqByClass',
     type AddGuarantees,
```
kozross, line 19: `type` is not necessary here: you only need it when exporting type _operators_.

src/Plutarch/Extra/Value.hs
```diff
@@ -1,6 +1,12 @@
+{-# LANGUAGE DuplicateRecordFields #-}
+{-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE OverloadedRecordDot #-}
+{-# LANGUAGE PartialTypeSignatures #-}
+{-# LANGUAGE PolyKinds #-}
 {-# LANGUAGE TypeApplications #-}
 
 module Plutarch.Extra.Value (
+  -- TODO: Label these imports better
```
kozross, line 9: I agree: no time like the present.

src/Plutarch/Extra/Value.hs
```diff
@@ -47,17 +79,14 @@ psingletonValue = phoistAcyclic $
...
 
-{- | Create a `PValue` that only contains specific amount tokens of the given assetclass.
-
-   @since 1.0.0
--}
+-- | Create a `PValue` that only contains specific amount tokens of the given assetclass.
 passetClassValue ::
     forall (s :: S) (keys :: KeyGuarantees) (amounts :: AmountGuarantees).
-    Term s (PAssetClass :--> PInteger :--> PValue keys amounts)
+    Term s (PAssetClassData :--> PInteger :--> PValue keys amounts)
```
kozross, line 85: Signature change, so requires `since` upgrade and a changelog entry.

src/Plutarch/Extra/Value.hs
```diff
@@ -109,10 +138,10 @@ psymbolValueOf =
    @since 1.1.0
 -}
 passetClassValueOf' ::
-    forall (keys :: KeyGuarantees) (amounts :: AmountGuarantees) (s :: S).
-    AssetClass ->
+    forall (tag :: Symbol) (keys :: KeyGuarantees) (amounts :: AmountGuarantees) (s :: S).
```
kozross, line 141: Same.

src/Plutarch/Extra/Value.hs
```diff
@@ -144,8 +173,8 @@ pgeqBySymbol =
    @since 1.1.0
 -}
 pgeqByClass' ::
-    forall (keys :: KeyGuarantees) (amounts :: AmountGuarantees) (s :: S).
-    AssetClass ->
+    forall (tag :: Symbol) (keys :: KeyGuarantees) (amounts :: AmountGuarantees) (s :: S).
```
kozross, line 176: Same.

src/Plutarch/Extra/Value.hs
```diff
@@ -170,3 +199,568 @@ phasOnlyOneTokenOfCurrencySymbol = phoistAcyclic $
     plam $ \cs vs ->
         psymbolValueOf # cs # vs #== 1
             #&& (plength #$ pto $ pto $ pto vs) #== 1
+
+-- | Elimator for PValueMap
```
kozross, line 203: `since` and quote-wrap for Haddock links.

src/Plutarch/Extra/Value.hs
```diff
@@ -170,3 +199,568 @@ phasOnlyOneTokenOfCurrencySymbol = phoistAcyclic $
...
+                                  )
+                                # symbolMaps
+                            )
+                        )
+                )
+                (self # symbolMaps)
+                (pto $ pfromData $ psndBuiltin # symbolMap)
+        )
+
+-- | Elimator for PValueMap
```
kozross, line 274: Same as above.

src/Plutarch/Extra/Value.hs
```diff
@@ -170,3 +199,568 @@ phasOnlyOneTokenOfCurrencySymbol = phoistAcyclic $
...
+                mnil
+                (pto $ pfromData $ psndBuiltin # symbolMap)
+        )
+        mnil
+
+{- |
+  Returns True if the provided comparison function returns True for all
+    entries in the first map with the same key in the second.
+-}
+pcmpMap ::
```
kozross, line 329: Same as above.

src/Plutarch/Extra/Value.hs
```diff
@@ -170,3 +199,568 @@ phasOnlyOneTokenOfCurrencySymbol = phoistAcyclic $
...
+
+{- |
+  Gets the first entry in the first item of the `PValue` mapping & returns the
+  rest of the PValue.
+
+  Fails if the PValue is empty. In cases where we know that a PValue contains
+  Ada, such as in the ScriptContext, then this will function will split the
+  Ada value - since the Ada entry comes first.
+-}
+psplitValue ::
```
kozross, line 380: Same as above.

src/Plutarch/Extra/Value.hs
```diff
@@ -170,3 +199,568 @@ phasOnlyOneTokenOfCurrencySymbol = phoistAcyclic $
...
+                        pto $
+                            pfromData
+                                (psndBuiltin # (phead # vList))
+                    )
+                    (pcon $ PValue $ pcon $ PMap $ ptail # vList)
+
+{- |
+  Returns True if all entries in the first Value present in the second Value
+  are >=.
+-}
```
kozross, line 414: Same as above.

src/Plutarch/Extra/Value.hs
```diff
@@ -170,3 +199,568 @@ phasOnlyOneTokenOfCurrencySymbol = phoistAcyclic $
...
+        )
+pgeqValue = phoistAcyclic $
+    plam $ \x y ->
+        x #== y
+            #|| pcmpMap # pgeqValueEntry # pto (pfromData x) # pto (pfromData y)
+
+{- |
+  Returns True if all entries in the first Map are >= the values in the second.
+-}
+pgeqValueEntry ::
```
kozross, line 431: Same as above.

src/Plutarch/Extra/Value.hs
```diff
@@ -170,3 +199,568 @@ phasOnlyOneTokenOfCurrencySymbol = phoistAcyclic $
...
+pgeqValueEntry = phoistAcyclic $
+    plam $ \_sym x y ->
+        x #== y
+            #|| (pcmpMap # plam (const pgeqIntegerData) # pfromData x # pfromData y)
+  where
+    pgeqIntegerData = phoistAcyclic $ plam $ \a b -> pfromData b #<= pfromData a
+
+{- | `plookup` but also returning the tail of the list, and relying on
+ sorted order
+-}
```
kozross, line 449: This needs a more meaningful name I feel. As well as a `since` and quote-links for the Haddock.

src/Plutarch/Extra/Value.hs
```diff
@@ -170,3 +199,568 @@ phasOnlyOneTokenOfCurrencySymbol = phoistAcyclic $
...
+                        pif
+                            (current #== needle)
+                            (pcon $ PJust $ pcon $ PPair (psndBuiltin # y) ysMap)
+                            (pif (needle #< current) (pcon PNothing) (self # ysMap))
+                )
+                (pcon PNothing)
+                (pto xs)
+
+-- | Cons case for QTokenValue, with recursive fixpoint
+findValue ::
```
kozross, line 478: If this is internal, it needs a regular comment, not a Haddock.

src/Plutarch/Extra/Value.hs
```diff
@@ -170,3 +199,568 @@ phasOnlyOneTokenOfCurrencySymbol = phoistAcyclic $
...
+mkSingleValue' (AssetClass sym tk) =
+    phoistAcyclic $ mkSingleValue # pconstantData sym # pconstantData tk
+
+-- TODO: Remove me, once Seungheon's PR lands
+pfromList ::
+    forall (list :: (S -> Type) -> S -> Type) (a :: S -> Type) (s :: S).
+    (PIsListLike list a) =>
+    [Term s a] ->
+    Term s (list a)
+pfromList = foldr (\x xs -> pcons # x # xs) pnil
```
kozross, line 766: Don't we have something like this already in `Plutarch.Extra.List`?

src/Plutarch/Orphans.hs
```diff
@@ -103,6 +108,22 @@ instance PTryFrom PData (PAsData PScriptHash) where
...
+    toBuiltinData rat =
+        BuiltinData $
+            List
+                [ I $ numerator rat
+                , I $ denominator rat
+                ]
+
+instance FromData (Ratio Integer) where
+    fromBuiltinData (BuiltinData (List [I num, I denom])) =
+        pure $ num % if num == 0 then 1 else denom
```
kozross, line 124: Why is the `if` necessary here?

src/Plutarch/Orphans.hs
```diff
@@ -103,6 +108,22 @@ instance PTryFrom PData (PAsData PScriptHash) where
...
+instance ToData (Ratio Integer) where
+    toBuiltinData rat =
+        BuiltinData $
+            List
+                [ I $ numerator rat
+                , I $ denominator rat
+                ]
+
+instance FromData (Ratio Integer) where
+    fromBuiltinData (BuiltinData (List [I num, I denom])) =
```
kozross, line 123: Nitpick, but I feel a `\case` would be better here.

kozross (changes requested):

SeungheonOh:
Since `AssetClass` and `AssetClassData` (the Hask) are identical, I think one can be just a newtype wrapper. (I doubt you can have one hask representation for two Pluts) It'd make conversion more easier as we can just coerce them.

src/Plutarch/Extra/AssetClass.hs
peter-mlabs: The old asset class was untagged, so I'm assuming this would make a difference?

peter-mlabs: What do you think?

src/Plutarch/Extra/AssetClass.hs
peter-mlabs: Good call. These were qualified downstream, but _if_ there was a conflict, there isn't any more.

src/Plutarch/Extra/ExchangeRate.hs
peter-mlabs: As an example: `exchangeFromTruncate` takes an exchange rate from `1.00 ADA :> 2.00 USD` and an `2.00 ADA` value, and returns `4.00 USD`. 

`echangeToTruncate` takes an exchange rate `1.00 ADA :> 2.00 USD` and a `4.00 USD` value, and returns `2.00 ADA`. 

Am I missing something else?

src/Plutarch/Extra/ExchangeRate.hs
peter-mlabs: yep

src/Plutarch/Extra/Rational.hs
peter-mlabs: Fixed

src/Plutarch/Extra/Rational.hs
peter-mlabs: Fixed the since.


RE argument order: I think the order of the comment is confusing (which I've fixed), but I think the actual order of the function makes sense. I think of things like `4 * (1/2)` or `4 / (1/2)` rather than the other way around.

peter-mlabs:
Addressed comments on everything besides the `Value` module. Will check the rest later.

src/Plutarch/Extra/Value.hs
peter-mlabs: Did a first pass here going by the comments and function names. May see some better opportunities for organization once I start looking at the actual functions themselves.

peter-mlabs:
I think the `PValue` comparison operations should have some more documentation. `Value` itself doesn't have an `Ord` instance, so we should probably do a [named chunk](https://haskell-haddock.readthedocs.io/en/latest/markup.html#named-chunks-of-documentation) explaining what ordering is used.

src/Plutarch/Extra/Value.hs
peter-mlabs: Doesn't seem like it. @SeungheonOh, I know you mentioned you'd add this -- is there anything special you wanted to do here (optimizations, etc?) or is it fine to just yeet in `Plutarch.Extra.List` as is?

SeungheonOh: I don't see any possible optimization here.

src/Plutarch/Orphans.hs
peter-mlabs: [See here](https://liqwidlabs.slack.com/archives/C01R16YV5EV/p1663779182508699?thread_ts=1663778686.919799&cid=C01R16YV5EV).

Short version: this may or may not still be necessary. We should check, and add to the documentation with the results.

peter-mlabs:
Note to self: need to move over the tests from Liqwid

src/Plutarch/Extra/AssetClass.hs
```diff
@@ -1,97 +1,362 @@
...
+          FromJSONKey
+        , -- | @since 3.8.0
+          ToJSONKey
+        )
+
+{- | A Scott-encoded Plutarch equivalent to 'AssetClass'.
+
+ @since 3.8.0
+-}
+data PAssetClass (unit :: Symbol) (s :: S) = PAssetClass
```
SeungheonOh, line 105: `unit` is called `tag` on `AssetClass`. we need consistency

src/Plutarch/Extra/AssetClass.hs
```diff
@@ -1,97 +1,362 @@
...
+ptoScottEncoding ::
+    forall (tag :: Symbol) (s :: S).
+    Term
+        s
+        ( PAsData PAssetClassData
+            :--> PAssetClass tag
+        )
+ptoScottEncoding = phoistAcyclic $
+    plam $ \cls ->
+        pletFields @["symbol", "name"] (pto cls) $
```
SeungheonOh, line 312: this `pto` can go I think. `PAssetClassData` implements `PDataFields`.

src/Plutarch/Extra/ExchangeRate.hs
```diff
@@ -0,0 +1,86 @@
...
+    forall (a :: Symbol) (b :: Symbol) (s :: S).
+    Term
+        s
+        ( PTagged (a :> b) PRational :--> PTagged b PInteger
+            :--> PTagged a PRational
+        )
+exchangeTo =
+    phoistAcyclic $
+        plam $ \ex x ->
+            ppure #$ divRational # pto x # pto ex
```
SeungheonOh, line 86: `untag` or `pextract`

src/Plutarch/Extra/ExchangeRate.hs
```diff
@@ -0,0 +1,86 @@
...
+exchangeFrom ::
+    forall (a :: Symbol) (b :: Symbol) (s :: S).
+    Term
+        s
+        ( PTagged (a :> b) PRational :--> PTagged a PInteger
+            :--> PTagged b PRational
+        )
+exchangeFrom =
+    phoistAcyclic $
+        plam $ \ex x -> ppure #$ mulRational # pto x # pto ex
```
SeungheonOh, line 70: `puntag` or `pextract`

src/Plutarch/Extra/ExchangeRate.hs
```diff
@@ -0,0 +1,86 @@
...
+exchangeToTruncate ::
+    forall (a :: Symbol) (b :: Symbol) (s :: S).
+    Term
+        s
+        ( PTagged (a :> b) PRational :--> PTagged b PInteger
+            :--> PTagged a PInteger
+        )
+exchangeToTruncate =
+    phoistAcyclic $
+        plam $ \ex x -> ppure #$ divTruncate # pto ex # pto x
```
SeungheonOh, line 54: `puntag` or `pextract`

src/Plutarch/Extra/Rational.hs
```diff
@@ -0,0 +1,133 @@
...
+-- | `plift` for Tagged Rationals (kind polymorphic)
+pliftTaggedRational ::
+    forall k (tag :: k).
+    HasCallStack =>
+    ClosedTerm (PTagged tag PRational) ->
+    Tagged tag Rational
+pliftTaggedRational term =
+    fromJust $
+        PlutusTx.fromData $
+            plift (pforgetData $ pdata term)
```
SeungheonOh, line 133: Can we include this as well. I remember seeing this function used quite extensively on liqwid-onchain.
https://github.com/mlabs-haskell/liqwid-onchain/blob/f107977798bbc6a65d67c2229fe6069a87d2e0f0/liqwid-onchain/Liqwid/Plutarch.hs#L522

src/Plutarch/Extra/Value.hs
```diff
@@ -1,92 +1,271 @@
...
+-}
+passetClassValueOf ::
+    forall
+        (key :: KeyGuarantees)
+        (amount :: AmountGuarantees)
+        (unit :: Symbol)
+        (s :: S).
+    Term s (PAssetClass unit :--> PValue key amount :--> PTagged unit PInteger)
+passetClassValueOf = phoistAcyclic $
+    plam $ \cls val -> unTermCont $ do
```
SeungheonOh, line 183: we can remove `TermCont` here.

SeungheonOh (commented):
Looks good overall. I'm not confident in `Value` module, but other modules looks nice.

SeungheonOh (approved):
LGTM.

kozross (approved):

peter-mlabs:
Merged in main, hopefully this will be good now

kozross (approved):

PR #88, opened 2022-09-25
Ensure fourmolu settings are followed

kozross:
This adds a `fourmolu.yaml`, and reformats everything to be compliant with it.


SeungheonOh (approved):
LGTM

PR #96, opened 2022-10-05
Remove SOP requirement

kozross:
This eliminates any need for us to ever derive SOP `Generic` again. Unfortunately, we can't be rid of that dependency in LPE itself, but now, none of our downstream consumers should need it.

Additionally, this adds significant clarifications to the original code, as well as additional compile-time safety checks. For example, `pconstantToRepr` for the 'producty' representation derivation no longer has any possibility of runtime failure, making it guaranteed total.

There's still work outstanding, and the module structure will likely change, hence the draft status.

### DIFF
--- a/src/Plutarch/Extra/PlutusType.hs
+++ b/src/Plutarch/Extra/PlutusType.hs
@@ -1,61 +1,214 @@
-{-# LANGUAGE UndecidableInstances #-}
 {-# LANGUAGE PolyKinds #-}
+{-# LANGUAGE QuantifiedConstraints #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE UndecidableSuperClasses #-}
 
 module Plutarch.Extra.PlutusType (
-  PlutusTypeDataList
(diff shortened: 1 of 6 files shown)


src/Plutarch/Extra/PlutusType.hs
```diff
@@ -0,0 +1,250 @@
...
+ - Both the Haskell and Plutarch type must implement 'Generic'.
+ - The Haskell and Plutarch types must have the same number of fields.
+ - The Haskell and Plutarch types must have equivalent types in each position.
+
+ If you get weird errors, check that all the above hold. Also, please report
+ it to us, so we can get better error messages!
+
+ @since 3.10.0
+-}
+data PlutusTypeRecord
```
SeungheonOh, line 80: This is a much better name :)

emiflake:
This repository has been archived. This PR should be recreated in the [liqwid-libs](https://github.com/liqwid-Labs/liqwid-libs) repo if it's still important. Closing.

PR #100, opened 2022-10-12
New FixedDecimal

SeungheonOh:
We should deprecate `P.E.Fixed` in near future.


src/Plutarch/Extra/Fixed.hs
```diff
@@ -0,0 +1,130 @@
...
+import Plutarch.Extra.Function (pflip)
+import Plutarch.Num (PNum (pfromInteger, (#*)))
+import qualified Plutarch.Numeric.Additive as A (
+  AdditiveMonoid (zero),
+  AdditiveSemigroup ((+)),
+ )
+import Plutarch.TryFrom (PTryFrom (PTryFromExcess, ptryFrom'))
+import Plutarch.Unsafe (punsafeCoerce)
+
+{- | Fixed width decimal. Decimal point will be given through typelit.
```
cjay, line 29: Denominator instead of decimal point.

src/Plutarch/Extra/Fixed.hs
```diff
@@ -0,0 +1,130 @@
...
+-- | @since 3.11.0
+instance KnownNat u => A.AdditiveMonoid (Term s (PFixed u)) where
+  zero = pcon . PFixed $ pconstant 0
+
+{- | Convert given decimal into Ada value. Input should be Ada value with decimals; outputs
+ will be lovelace values in integer.
+
+ @since 3.9.0
+-}
+decimalToAdaValue ::
```
cjay, line 100: Should be renamed.

src/Plutarch/Extra/Fixed.hs
```diff
@@ -0,0 +1,130 @@
...
+
+-- | @since 3.11.0
+instance KnownNat u => A.AdditiveSemigroup (Term s (PFixed u)) where
+  (+) = (+)
+
+-- | @since 3.11.0
+instance KnownNat u => A.AdditiveMonoid (Term s (PFixed u)) where
+  zero = pcon . PFixed $ pconstant 0
+
+{- | Convert given decimal into Ada value. Input should be Ada value with decimals; outputs
```
cjay, line 95: "fixed" instead of "decimal".

cjay (approved):
Looks good.

kozross (changes requested):

SeungheonOh:
Thanks for the review! `Plutarch.Extra.FixedDecimal` was the newly added module and `Plutarch.Extra.Fixed` is the old one that was remained since it's not really `decimal`. I haven't got any comment of the `FixedDecimal` so I am just making sure it's getting review as well. Thanks.

kozross (approved):

cjay:
Hmm I think `(#*)` in `PFixed` should also use `pto` instead of `punsafeCoerce`, like `PFixedDecimal` does. @kozross do you know if a bit-shift will be cheaper than a division on big integers in the Plutus cost model? If yes, maybe we should add a `FixedBinary`, for power-of-2 denominators, so we can get a free performance boost in the future.

cjay:
And should `PFixedDecimal` really have a `PIntegral` instance? It looks like it will behave correctly, but it's unexpected :thinking:

cjay:
I just found that there is a `Data.Fixed` in base, never heard of that before. We sholud try to be compatible with the way it specifies denominators, to be able to convert to/from Haskell values. It leaves the base open, but defines stuff like `E6` which will result in having a denominator of one million. Hmm I hope Purescript has something comparable, for offchain stuff.

cjay:
Though this would be incompatible with binary fractions to allow for bit shifting. So maybe we shouldn't do this. Hmm.

SeungheonOh:
`Data.Fixed` in the base is in base-10, so technically it should be represented with `Plutarch.Extra.PFixedDecimal` (the one that we changed). I'm not sure if this is the best idea because naming is quite confusing this way

kozross:
@cjay In theory, assuming my design and implementation was used, a shift would be faster. Haskell by default does not do shift conversion, even on constants, for `Integer`s: I established this with benchmarks, including for bitwise primops. However, the design and implementation I wrote for bitwise primops keeps a 'conversion firewall' between numbers and blobs of bits, so any such advantage would probably be lost for single operations.

PR #102, opened 2022-10-13
Rational: One less comparison, if the denominator is positive already

cjay:
Previously, the use of `ptryPositive` in `#%` caused a duplicate `#<=` comparison. This change reduces the number of comparisons in the positive case. It could also be changed to optimize the negative case instead, but I'd expect that to be less useful. One of the branches has to do the explicit check for 0 due to `PPositive` being used.

### DIFF
--- a/src/Plutarch/Extra/Rational.hs
+++ b/src/Plutarch/Extra/Rational.hs
@@ -21,7 +21,8 @@ import Plutarch.Builtin (pforgetData)
 import Plutarch.Extra.Tagged (PTagged)
 import "plutarch-extra" Plutarch.Extra.TermCont (pmatchC)
 import Plutarch.Orphans ()
-import Plutarch.Positive (ptryPositive)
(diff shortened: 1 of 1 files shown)


kozross (approved):
Could this use the three-way-comparison operations in `Extra.Ord`?

cjay:
Which function? I don't see one that offers three way branching. The flexibility of `PComparator` is also not needed here.

kozross:
I agree, the generality of `PComparator` isn't needed. We have `pcompareBy` which returns a `POrdering`, which can be used for a three-way branch.

cjay:
Creating that intermediate value would reduce efficiency, doesn't seem like a good idea.

kozross:
Fair enough - I figured as much.

cjay:
Do I need to make a change log entry for this?

kozross:
No - it's a purely implementation change, so it doesn't even need a version bump.

PR #103, opened 2022-10-14
made `withStateThread` parameterized at Plutarch

SeungheonOh:


peter-mlabs (changes requested):
needs haddocks, sinces, changelog, and cabal version bump, but otherwise looks fine

kozross (changes requested):
The naming seems a bit off. Also, the docs should indicate clearly what the two different functions differ in.

Also what @peter-mlabs said.

peter-mlabs (approved):

kozross (approved):

PR #105, opened 2022-10-20
Improve ergonomics of `AssetClass` APIs

chfanghr:
Use cases: 
- https://github.com/Liqwid-Labs/agora/pull/197/files/467c4abcd1fd3a66606caaa39c9b06c4eae8ddd7#diff-f8c93205d1ca2676edb4296f2befc56f28a2296a3abf31fb2ef1029dccb2d186
- https://github.com/Liqwid-Labs/agora/pull/197/files/467c4abcd1fd3a66606caaa39c9b06c4eae8ddd7#diff-f23b92895ae72b59d05759224bed0614f448c7cd083c7fc6985393a2c92f92e9


src/Plutarch/Extra/AssetClass.hs
```diff
@@ -150,9 +158,9 @@ passetClass = phoistAcyclic $
   plam $ \sym tk ->
     pcon $ PAssetClass (pdata sym) (pdata tk)
 
--- | @since 3.10.0
+-- | @since 3.10.3
 passetClassT ::
-  forall (unit :: Symbol) (s :: S).
```
SeungheonOh, line 155: these can be given `(unit :: k)` maybe? or `{k} (unit :: k)` if the first one doesn't work.

SeungheonOh (changes requested):
Looks good. I think we can give an explicit kind sigs.

src/Plutarch/Extra/AssetClass.hs
```diff
@@ -150,9 +158,9 @@ passetClass = phoistAcyclic $
...
+  forall {k} (unit :: k) (s :: S).
```
kozross, line 163: I'd rather an explicit than implicit argument here.

src/Plutarch/Extra/AssetClass.hs
```diff
@@ -173,9 +181,9 @@ pconstantClass ac =
       (pconstantData $ view #symbol ac)
       (pconstantData $ view #name ac)
 
--- | @since 3.10.0
+-- | @since 3.10.3
 pconstantClassT ::
-  forall (unit :: Symbol) (s :: S).
+  forall {k} (unit :: k) (s :: S).
```
kozross, line 186: Same here.

src/Plutarch/Extra/AssetClass.hs
```diff
@@ -192,10 +200,10 @@ psymbolAssetClass = phoistAcyclic $
     pcon $ PAssetClass (pdata sym) emptyTokenNameData
 
 {- | Tagged version of `psymbolAssetClass`
- @since 3.10.0
+ @since 3.10.3
 -}
 psymbolAssetClassT ::
-  forall (unit :: Symbol) (s :: S).
+  forall {k} (unit :: k) (s :: S).
```
kozross, line 206: Also here.

src/Plutarch/Extra/AssetClass.hs
```diff
@@ -304,21 +312,21 @@ passetClassData = phoistAcyclic $
           .& #name .= pdata tk
       )
 
--- | @since 3.10.0
+-- | @since 3.10.3
 passetClassDataT ::
-  forall (unit :: Symbol) (s :: S).
+  forall {k} (unit :: k) (s :: S).
```
kozross, line 317: And here.

src/Plutarch/Extra/AssetClass.hs
```diff
@@ -330,40 +338,39 @@ ptoScottEncoding = phoistAcyclic $
...
     fn #$ ptoScottEncoding # cls
 
 {- | Version of 'assetClassValue' for tagged 'AssetClass' and 'Tagged'.
 
- @since 3.9.0
+ @since 3.10.3
 -}
 assetClassValue ::
-  forall (unit :: Symbol).
+  forall {k} (unit :: k).
```
kozross, line 373: And here.

kozross (changes requested):
I feel this is a major bump: you are changing signatures pretty fundamentally. However, I don't really know if the `PAsData` being removed would cause existing code using these to fail to compile or not.

src/Plutarch/Extra/AssetClass.hs
SeungheonOh: We moved away from `Symbol` to use type tags as well. How would an explicit polykind look like?

kozross: `forall (k :: Type) (unit :: k) ...`

`Type :: Type` and all that.

kozross: It might be less convenient in that ordering, so maybe `forall (unit :: k) (k :: Type) ...` would be better.

SeungheonOh: :bulb: Ah, that makes sense.

chfanghr: `forall (unit :: k) (k :: Type)` won't work cause GHC can't find `k`.

chfanghr: I guess we can make `k` inference only?

emiflake: Inference only for `k` makes sense to me here. We definitely want to specify the unit first.

chfanghr:
Please don't merge yet. Removing `PAsData` seems to be problematic.

SeungheonOh:
For Ply, I think we should use toData function to reduce confusion. 

https://github.com/Liqwid-Labs/oracle-onchain/blob/main/src/Liqwid/Oracle/Linker.hs

Actually, I can set up a deriving strategy for this, but that will work for now

src/Plutarch/Extra/AssetClass.hs
peter-mlabs: curious what the practical difference is between `Symbol` and just regular type tags. Obviously the latter is more polymorphic, but when would we want to _not_ use a `Symbol` in practice?

peter-mlabs (changes requested):
Not much to add except that you need a version bump in the cabal

src/Plutarch/Extra/AssetClass.hs
```diff
@@ -392,3 +399,18 @@ instance
   LabelOptic "name" k (Tagged tag AssetClass) (Tagged tag' AssetClass) a b
   where
   labelOptic = #unTagged %% #name
+
+----------------------------------------
+-- Ply instances
+
+-- | @since 3.10.3
+instance PlyArg AssetClass where
```
SeungheonOh, line 407:
```hs
instance PlyArg AssetClass where
  type UPLCRep AssetClass = [Plutus.Data]
  toBuiltinArg ac =
    case Plutus.toData @AssetClass ac of
      Plutus.List x -> x
      _ -> error "unreachable"
  toBuiltinArgData = Plutus.toData
```

Sorry, I should've explained it better. Basically, we don't want to expose actual logic here, so that when we change `AssetClass`, it doesn't break.

chfanghr: How about using `PlyArg` as the single source of truth so that we don't have to do deconstruction?

src/Plutarch/Extra/AssetClass.hs
chfanghr: See https://github.com/Liqwid-Labs/agora/blob/205241a7ad6b7f631aa670969cc25ef3128d09b8/agora/Agora/SafeMoney.hs

src/Plutarch/Extra/AssetClass.hs
SeungheonOh: the single source of truth should be the derivation result as we don't have control over that. I'm not sure where deconstruction came from 🤔. 

Also, most `PlyArg` instances in ply themselves are defined similar to this.

chfanghr: Good point.

SeungheonOh: Thanks!

peter-mlabs (approved):
LGTM; I'll wait for @kozross 's review and then we can merge

flake.nix
```diff
@@ -47,6 +48,8 @@
           "${inputs.plutarch-quickcheck}"
           "${inputs.plutarch-numeric}"
           "${inputs.plutarch-context-builder}"
+          "${inputs.ply}/ply-core"
```
SeungheonOh, line 51: I don't know if it's advisable, but I used `${inputs.liqwid-script-export.inputs.ply}/ply-{core|plutarch}` for other repositories.

SeungheonOh (commented):
Looks good. Thank you!

flake.nix
chfanghr: We have a cyclic dependency issue here.

SeungheonOh: yeah, oh no. Perhaps we just put all ply orphans into LSE?

SeungheonOh: We have very confusing orphan situation:

Agora and LPE both have orphans for Aeson instances. Propbably liqwid also has its own.
LSE requires LPE for all Aeson instances and Ply instances (if we decide to put it on LPE).

Since now, we don't need Aeson instances for individual onchain repo, we can put all Aeson instances and Ply instances into LSE. I think that will make LSE independent of LPE.

kozross (approved):

flake.nix
SeungheonOh: That makes sense. Let's just go with this. :+1:

SeungheonOh (approved):

PR #107, opened 2022-10-24
Compliance with new standards

kozross:
Since we now require `DuplicateRecordFields` and `UndecidableInstances` to be on everywhere, this PR sets them to be always-on, and removes any case of them being on per-module.


SeungheonOh (approved):
LGTM

PR #108, opened 2022-10-25
FixedDecimal Haskell version and missing instances

cjay:


peter-mlabs:
Does this supersede #100 ?

peter-mlabs:
Tagging Koz in as a reviewer because Seungheon is off this week

cjay:
Thanks. 

There is a possible optimization here for code that has to do multiple multiplications and divisions in a row: We could change the type-level exponent to Integer and use `KnownInt` from `typenums` to allow for exponents with any sign. Then, multiplications and divisions could update the type-level integer to push as many calculations as  possible off-chain. I planned to try this, but have put it on hold because I saw that it would require an external library (`typenums`).

cjay:
Testing with Natural, seems to work fine (but a bit less flexible since having to stay >= 0)

```haskell
emul :: forall (expA :: Natural) (expB :: Natural).
  FixedDecimal expA -> FixedDecimal expB -> FixedDecimal (expA + expB)
emul (FixedDecimal a) (FixedDecimal b) = FixedDecimal $ a * b

ediv :: forall (expA :: Natural) (expB :: Natural).
  FixedDecimal expA -> FixedDecimal expB -> FixedDecimal (expA - expB)
ediv (FixedDecimal a) (FixedDecimal b) = FixedDecimal $ a `div` b
```

The existing `pconvertExp` can then finalize the calculation with one division only, as required.

cjay:
Added them in. They should make some calculations in Liqwid more elegant and closer to the spec.

SeungheonOh:
https://github.com/Liqwid-Labs/liqwid-plutarch-extra/commit/1b0075d3bc3519d89f5af5d515f281594ad0b361

I found some unnecessary `pconstant`s and `ptoIntegers`.

cjay:
Good point. Feel free to merge, I guess the final thing is easier to review against staging in your original PR.

kozross (approved):

PR #109, opened 2022-10-25
QuickCheck support for AssetClass

kozross:
This adds a modifier newtype, similar to `GenCurrencySymbol`, for QuickCheck support for `AssetClass`. It allows control over whether you want the ADA asset class to be generatable or not.

### DIFF
--- a/src/Plutarch/Extra/AssetClass.hs
+++ b/src/Plutarch/Extra/AssetClass.hs
@@ -33,6 +33,10 @@ module Plutarch.Extra.AssetClass (
   ptoScottEncoding,
   pfromScottEncoding,
   pviaScottEncoding,
+
+  -- * Modifiers for QuickCheck
+  AdaACPresence (..),
+  GenAssetClass (..),
 ) where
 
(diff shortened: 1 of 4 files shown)


src/Plutarch/Extra/AssetClass.hs
```diff
@@ -411,3 +427,158 @@ instance PlyArg AssetClass where
 
 -- | @since 3.10.4
 type instance PlyArgOf PAssetClassData = AssetClass
+
+{- | Type-level marker to indicate whether a 'GenAssetClass' can have the ADA
+ 'AssetClass' inside it or not.
+
+ @since 3.11.1
+-}
+data AdaACPresence = WithAdaAC | WithoutAdaAC
```
peter-mlabs, line 436: Use `Class` instead of `AC`

peter-mlabs (approved):
Looks good, just need to change `AC` to `Class`.

peter-mlabs (approved):
LGTM

PR #111, opened 2022-10-27
Added integer power to Numeric, along with property test

cjay:
Upstreamed from Liqwid.

### DIFF
--- a/src/Plutarch/Extra/Numeric.hs
+++ b/src/Plutarch/Extra/Numeric.hs
@@ -1,7 +1,8 @@
 {-# LANGUAGE QuantifiedConstraints #-}
 
-module Plutarch.Extra.Numeric (peven) where
+module Plutarch.Extra.Numeric (peven, (#^)) where
 
+import Plutarch.Extra.TermCont (pletC)
 import Plutarch.Num (PNum)
 
 -- | @since 1.0.0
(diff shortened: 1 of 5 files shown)


kozross (changes requested):
Cabal file needs a version bump.

cjay:
Done, and added a newtype wrapper to prevent `PRational` reduction, since the original power function in Liqwid prevented reduction internally.

src/Plutarch/Extra/Numeric.hs
```diff
@@ -10,3 +11,46 @@ peven ::
...
+ the 'Plutarch.Extra.PRationalNoReduce' wrapper to prevent reducing after each
+ internal multiplication.
+
+ @since 3.12.2
+-}
+(#^) ::
+  forall (s :: S) (n :: S -> Type) (e :: S -> Type).
+  (PNum n, PNum e, PEq e, PIntegral e) =>
+  Term s n ->
+  Term s e ->
```
kozross, line 52: Could we use `PPositive` for `e` here?

kozross (approved):

src/Plutarch/Extra/Numeric.hs
cjay: Excluding 0 would restrict it too much, it would have to be `PNatural`. Should we make a `PNatural` in LPE?

kozross: Ah good point. I think for now we can leave it as `PIntegral`, as partiality in Plutarch isn't quite as bad as in Haskell.

Maybe `PNatural` would be a good thing for LPE in the future.

PR #112, opened 2022-10-27
FixedDecimal: Zero-cost conversions and optimization

cjay:
Forgot something. Also, most `pconstant` are back in, because type inference made the 10 a `Term`, so the `^` would create Plutarch-level calculations.


kozross (changes requested):
Cabal file needs a version bump.

cjay:
Done.

kozross (approved):

PR #116, opened 2022-11-01
Update liqwid-nix

kozross:
This upgrades us to the latest `liqwid-nix`. This requires some reformatting to work, due to both `fourmolu` and `cabal-fmt` being bumped: the change surface thus appears larger than it actually is.


SeungheonOh (approved):
LGTM, nothing to review tbh :)

PR #117, opened 2022-11-02
Upstream utilities from agora

chfanghr:


SeungheonOh:
first off, need to bump cabal version

src/Plutarch/Extra/Applicative.hs
```diff
@@ -236,3 +239,42 @@ punless ::
...
+    Term s (f a)
+
+-- | @since 3.14.1
+instance PAlternative PMaybe where
+  (#<|>) = phoistAcyclic $
+    plam $ \a b -> pmatch a $ \case
+      PNothing -> b
+      PJust _ -> a
+  pempty = pnothing
+
```
SeungheonOh, line 263: instance for `PEither` would be helpful as well.

chfanghr:
We should probably add a pr template.

src/Plutarch/Extra/List.hs
```diff
@@ -75,15 +89,30 @@ pmapMaybe ::
...
   (PElemConstraint ell a, PElemConstraint ell b, PListLike ell) =>
   Term s ((a :--> PMaybe b) :--> ell a :--> ell b)
-pmapMaybe = phoistAcyclic $ plam $ \f -> precList (go f) (const pnil)
+pmapMaybe = pmapMaybe'
+
+{- | A version of 'pmapMaybe' that can change the list type.
+
+     @since 3.14.1
+-}
+pmapMaybe' ::
```
SeungheonOh, line 98: This is neat. Actually, I think it can replace `pmapMaybe`. I don't think there will be any performance overhead.

src/Plutarch/Extra/List.hs
```diff
@@ -209,3 +238,132 @@ pfromList ::
   [Term s a] ->
   Term s (list a)
 pfromList = foldr (\x xs -> pcons # x # xs) pnil
+
+{- | Check if two lists are equal, given a custom equalator. The two list types
+     can be different.
+
+     @since 3.14.1
+-}
+plistEqualsBy ::
```
SeungheonOh, line 247: Maybe use `PComparator`, but it's really fine either way.

src/Plutarch/Extra/List.hs
```diff
@@ -209,3 +238,132 @@ pfromList ::
...
+      -- l1 is empty, so l2 should be empty as well.
+      (pnull # l2)
+      l1
+
+{- | / O(n) /. Remove the first occurance of a value from the given list. Return
+     nothing is said value is not in the list.
+
+     @since 3.14.1
+-}
+pdeleteBy ::
```
SeungheonOh, line 278: We should put first in the name to be clear. Perhaps, `pdeleteFirstBy`

src/Plutarch/Extra/List.hs
```diff
@@ -209,3 +238,132 @@ pfromList ::
...
+            t
+            (pcons # h #$ self # t)
+      )
+      (const $ ptraceError "Cannot delete element")
+
+{- | Special version of 'pdeleteBy', for types with 'POrd' instance.
+
+     @since 3.14.1
+-}
+pdelete ::
```
SeungheonOh, line 322: same as comment above `pdeleteFirst`

src/Plutarch/Extra/List.hs
```diff
@@ -209,3 +238,132 @@ pfromList ::
...
+    precList
+      (\_ _ t -> pnull # t)
+      (const $ pconstant False)
+
+{- | Extract a singeton value from the given list. Throw an error if the list
+     consists of zero or more than one elements.
+
+     @since 3.14.1
+-}
+pfromSingleton ::
```
SeungheonOh, line 353: give nonpartial version. like `pdeleteBy`

src/Plutarch/Extra/Ord.hs
```diff
@@ -1029,3 +1032,31 @@ pcmpValue self doCmp cmp k1 k2 v1 v2 kv1 kv2 = pmatch (pcompareBy # pfromOrd # k
...
+          ( \self h t ->
+              let ensureUniqueness =
+                    pif
+                      (eq # x # h)
+                      (ptraceError "inserted value already exists")
+                  next =
+                    pif
+                      (lt # x # h)
+                      (pcons # x #$ pcons # h # t)
+                      (pcons # h #$ self # t)
```
SeungheonOh, line 1059: It also respects the sorted-ness. Please add documentation.

src/Plutarch/Extra/ScriptContext.hs
```diff
@@ -439,3 +451,60 @@ pfindOwnInput = phoistAcyclic $
...
+
+    @since 3.14.1
+-}
+ptryFromRedeemer ::
+  forall (r :: PType) (s :: S).
+  (PTryFrom PData r) =>
+  Term
+    s
+    ( PScriptPurpose
+        :--> PMap 'Unsorted PScriptPurpose PRedeemer
```
SeungheonOh, line 503: It can accept Unsorted and Sorted.

src/Plutarch/Extra/Value.hs
```diff
@@ -778,3 +781,46 @@ matchOrTryRec requiredAsset = phoistAcyclic $
       )
       (const def)
       # inputpvalue
+
+{- | Get the negative and positive amount of a particular 'CurrencySymbol', and
+     return nothing if it doesn't exist in the value.
+
+     @since 3.14.1
+-}
+psymbolValueOf' ::
```
SeungheonOh, line 790: Looks good, but this is oddly specific :D

SeungheonOh (changes requested):

SeungheonOh:
LGTM, some comments.

src/Plutarch/Extra/List.hs
chfanghr: Changing the signature of `pmapMaybe` will make existing code uncompilable. I'm not sure people are okay with this.

src/Plutarch/Extra/List.hs
chfanghr: Hmmm, in agora's use case, the element might not have `POrd`.

src/Plutarch/Extra/List.hs
chfanghr: Good idea.

src/Plutarch/Extra/Value.hs
chfanghr: Haha, it is. It's efficient for that specific use case, though.

src/Plutarch/Extra/ScriptContext.hs
chfanghr: The intended use case is to pass in the `redeemers` field of `PTxInfo`, which is defined as:
```hs
"redeemers" := PMap 'Unsorted PScriptPurpose PRedeemer
```
I should mention it in the doc.

src/Plutarch/Extra/Value.hs
chfanghr: For context, see the discussion in https://github.com/Liqwid-Labs/agora/pull/202

src/Plutarch/Extra/Applicative.hs
chfanghr: I am not sure what's the identity of `PEither`.

chfanghr: Actually, `Either` doesn't have `Applicative` instance.

src/Plutarch/Extra/List.hs
kozross: I don't think this will happen: the worst it'll cause is some added type parameters to direct our choice of lists. Plus, major bump and all that - as long as you changelog it, I don't see the issue.

src/Plutarch/Extra/List.hs
kozross: You can construct `PComparator`s without needing an `Ord` instance. I agree with @SeungheonOh here: `PComparator`s at least force you to be honest.

src/Plutarch/Extra/List.hs
```diff
@@ -209,3 +239,154 @@ pfromList ::
...
+pdeleteFirstBy = phoistAcyclic $
+  plam $ \f' x -> plet (f' # x) $ \f ->
+    precList
+      ( \self h t ->
+          pif
+            (f # h)
+            (pjust # t)
+            (pfmap # (pcons # h) # (self # t))
+      )
+      (const pnothing)
```
kozross, line 295: I assume the function argument is meant to be an equality check? If so, it should be documented; if not, an explanation of what it is would help a lot.

src/Plutarch/Extra/List.hs
```diff
@@ -209,3 +239,154 @@ pfromList ::
...
+pdeleteFirstBy' = phoistAcyclic $
+  plam $ \f' x -> plet (f' # x) $ \f ->
+    precList
+      ( \self h t ->
+          pif
+            (f # h)
+            t
+            (pcons # h #$ self # t)
+      )
+      (const $ ptraceError "Cannot delete element")
```
kozross, line 317: The convention for naming variants that crash instead of returning `PMaybe` or similar is to use `try`. So in this case, it'd be `ptryDeleteFirstBy`.

src/Plutarch/Extra/List.hs
```diff
@@ -209,3 +239,154 @@ pfromList ::
...
+     @since 3.14.1
+-}
+pdeleteFirst ::
+  forall
+    (a :: S -> Type)
+    (list :: (S -> Type) -> (S -> Type))
+    (s :: S).
+  (PEq a, PIsListLike list a) =>
+  Term s (a :--> list a :--> PMaybe (list a))
+pdeleteFirst = phoistAcyclic $ pdeleteFirstBy # plam (#==)
```
kozross, line 330: The documentation disagrees with the implementation: you only require `PEq`.

src/Plutarch/Extra/List.hs
```diff
@@ -209,3 +239,154 @@ pfromList ::
...
+pfromSingleton' =
+  phoistAcyclic $
+    precList
+      ( \_ h t ->
+          pif
+            (pnull # t)
+            h
+            (ptraceError "More than one element")
+      )
+      (const $ ptraceError "Empty list")
```
kozross, line 392: As previously: `ptryFromSingleton`.

kozross (changes requested):

src/Plutarch/Extra/Applicative.hs
kozross: [Yes it does](https://hackage.haskell.org/package/base-4.17.0.0/docs/Data-Either.html#t:Either).

kozross: I guess you meant `Alternative`? Then indeed not, but it [_can_ be an `Alt`](https://hackage.haskell.org/package/semigroupoids-5.3.7/docs/Data-Functor-Alt.html#t:Alt).

chfanghr: Added `instance PAlt (PEither a)`

kozross (changes requested):

src/Plutarch/Extra/Bool.hs
```diff
@@ -40,3 +41,19 @@ pcompare t1 t2 ifLT ifEQ ifGT =
 -}
 pcond :: forall (s :: S) (a :: S -> Type). [Term s a -> Term s a] -> Term s a -> Term s a
 pcond = appEndo . foldMap Endo
+
+{- | If the condition evaluated to true, return the third argument. Otherwise error
+     out with the error message.
+
+     @since 3.14.1
+-}
+passert ::
```
peter-mlabs, line 50: always fascinating to me which functions we were missing. I could have _sworn_ there was a function that did this already, but looking through hoogle it appears there wasn't!

src/Plutarch/Extra/List.hs
```diff
@@ -209,3 +228,154 @@ pfromList ::
...
+            -- l2 is empty, but l1 is not.
+            (pconstant False)
+            l2
+      )
+      -- l1 is empty, so l2 should be empty as well.
+      (pnull # l2)
+      l1
+
+{- | / O(n) /. Remove the first occurance of a value from the given list. Return
+     nothing is said value is not in the list.
```
peter-mlabs, line 264: "Return nothing _if_ said value (...)"

src/Plutarch/Extra/List.hs
peter-mlabs: Actually, I think it would probably be more useful to have this be `Term s ((a :--> PBool) :--> list a :--> PMaybe (list a))`,
which makes it more clear that its deleting the first element that satisfies any predicate.

What do you think?

peter-mlabs: likewise for `ptryFirstDeleteBy`

peter-mlabs (approved):
Looks great overall, thanks for these Connor! 

I have a few minor comments, but nothing substantial to add to the discussion. Feel free to merge when you're ready (or ping me if CI is still broken at that time)

src/Plutarch/Extra/List.hs
chfanghr: Good idea. Will make the change.

src/Plutarch/Extra/Bool.hs
chfanghr: There's always something missing. Just added a few as I drafted the pr on agora side: https://github.com/Liqwid-Labs/agora/blob/fc695e46c7fa9bf74ee2b57317188f0e0c407c41/agora/Agora/Utils.hs

SeungheonOh (approved):
LGTM. Thanks!

kozross (approved):

PR #119, opened 2022-11-08
inspect and inspects

kozross:
This saves typing time by providing a (highly polymorphic) helper `inspect`, which combines `view` and `asks`. Also provided is `inspects`, which parallels `views`.


SeungheonOh (approved):
Looks helpful. LGTM

PR #124, opened 2022-11-10
pextendedAssetClassValueOf'

kozross:
This adds a parallel to `passetClassValueOf'` for `ExtendedAssetClass`.

### DIFF
--- a/liqwid-plutarch-extra.cabal
+++ b/liqwid-plutarch-extra.cabal
@@ -1,6 +1,6 @@
 cabal-version:      3.0
 name:               liqwid-plutarch-extra
-version:            3.14.6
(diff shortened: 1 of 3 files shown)


liqwid-plutarch-extra.cabal
```diff
@@ -1,6 +1,6 @@
 cabal-version:      3.0
 name:               liqwid-plutarch-extra
-version:            3.14.6
```
SeungheonOh, line 3: this should be `3.14.6`

SeungheonOh (changes requested):
incorrect version. Other things look good!

SeungheonOh (approved):
LGTM

PR #125, opened 2022-11-13
More ExtendedAssetClass helpers

kozross:
This adds two constants (for Haskell and Plutarch) representing the ADA asset class, as well as a predicate for verifying whether an arbitrary `ExtendedAssetClass` is the ADA one. This also 'burns in' the decision that the ADA class has a fixed, rather than arbitrary, token name, as per @t1lde's suggestion.


SeungheonOh (approved):
Looks good. Maybe `pisExtendedAdaClass` is also helpful.

kozross:
Yeah, I might as well throw that in.

PR #126, opened 2022-11-14
HasLabelled

kozross:
Amidst my work on `liqwid-onchain` to use `ExtendedAssetClass`, I found myself needing setter capability. However, `HasLabelledGetters` isn't powerful enough to describe that we have _lenses_ on a particular set of fields, only getters.

This PR provides the ability to 'collectively constrain' any number of fields to any optic of our choice, using a type family `HasLabelled`. Furthermore, we provide a backwards-compatible version of `HasLabelledGetters` based on this new design. This has been marked as deprecated for now, but at least it won't break anything.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 This format is based on [Keep A Changelog](https://keepachangelog.com/en/1.0.0).
 
+## 3.15.2 -- 2022-11-15
+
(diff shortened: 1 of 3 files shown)


SeungheonOh (approved):
Neat. This is good. ~~except ofc the type synonyms~~. LGTM.

PR #127, opened 2022-11-14
guarantee and guarantees

kozross:
This provides `guarantee` and `guarantees`, which are variants of `preview` and `previews` which allow giving a default, similar to `fromMaybe`.


SeungheonOh (approved):
LGTM

PR #128, opened 2022-11-15
Restructure tests

SeungheonOh:


liqwid-plutarch-extra.cabal
```diff
@@ -160,10 +160,16 @@ library
...
-test-suite properties
+test-suite test
   import:         common-lang
   type:           exitcode-stdio-1.0
   main-is:        Main.hs
+  other-modules:
+    Properties.FunctorProp
+    Properties.NumericProp
+    Properties.OrdProp
+    Properties.TraversableProp
```
kozross, line 171: The `Prop` suffix is redundant. I'd even go so far as not bothering to have more than a single-layer hierarchy: we're not exporting this publically, no chance of clashing with anything.

test/Properties/NumericProp.hs
```diff
@@ -0,0 +1,29 @@
...
+      ]
+  where
+    go :: QuickCheckTests -> QuickCheckTests
+    go = max 1000
+
+-- Properties
+
+propPowInt :: NonNegative Integer -> NonNegative Integer -> Property
+propPowInt (NonNegative n) (NonNegative i) =
+  (n ^ i) === plift (pconstant n #^ pconstant i)
```
kozross, line 29: If we want to test exponentiation, that has laws of its own:

* `x ^ i * x ^ j = x ^ (i + j)`
* `(x ^ y) ^ z = x ^ (y * z)`
* `x ^ 1 = x`
* `x ^ 0 = 1`

These are actually demonstrating a semiring-like structure on exponentiation, hence the two identities (additive and multiplicative), and showing that we have a 'faithful' translation.

kozross (changes requested):

kozross (changes requested):

kozross (approved):

PR #129, opened 2022-11-16
eqClasses, unsafeToAssetClass

kozross:
This adds two useful `ExtendedAssetClass` functions. Both were originally defined for Plutarch: these are their Haskell equivalents.


SeungheonOh (approved):
LGTM, they are all Haskell version of existing Plutarch one!

PR #134, opened 2022-11-27
Expose `UD'`

t4ccer:
I need this exposed to have encoding of polymorphic types working


t4ccer:
I don't need this anymore

PR #136, opened 2022-11-30
Rename utilities in `Plutarch.Extra.Time`. 

chfanghr:
According to the agora audit team:

```
`PCurrentTime` naming in Liqwid-Plutarch-Extra can be confusing because it refers to the whole transaction validity range. The validity range can be very long with the actual current time somewhere in the long range.
```

The purpose of this pr is to eliminate the possible confusion from the naming scheme and thus doesn't contain any actual code changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,21 @@
 
 This format is based on [Keep A Changelog](https://keepachangelog.com/en/1.0.0).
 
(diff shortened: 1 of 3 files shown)


peter-mlabs (approved):
All reasonable! thanks

SeungheonOh (approved):
Looks good

PR #137, opened 2022-12-02
Fix bug in state token

peter-mlabs:
Fixes a bug noted by @chfanghr [here](https://github.com/Liqwid-Labs/agora-pro/pull/23#discussion_r1037768988)

### DIFF
--- a/src/Plutarch/Extra/StateThread.hs
+++ b/src/Plutarch/Extra/StateThread.hs
@@ -4,6 +4,7 @@ module Plutarch.Extra.StateThread (
 ) where
 
 import Plutarch.Api.V1 (PCurrencySymbol, PValue)
+import Plutarch.Api.V1.AssocMap (plookup)
 import Plutarch.Api.V1.Value (pvalueOf)
 import Plutarch.Api.V2 (
   AmountGuarantees,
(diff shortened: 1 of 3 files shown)


src/Plutarch/Extra/StateThread.hs
```diff
@@ -59,7 +62,13 @@ uniqueStateTokenMint ::
   Term s (PValue keys amounts) ->
   Term s PBool
 uniqueStateTokenMint thisPolicy mint =
-  pvalueOf # mint # thisPolicy # pconstant "" #== pconstant 1
+  pisSingleton
```
chfanghr, line 65: You can use `ptryFromSingleton` and check the token name - value pair so the the second check can be avoided.

chfanghr (approved):
This should do the trick. Thanks for the fix!

src/Plutarch/Extra/StateThread.hs
```diff
@@ -59,7 +62,13 @@ uniqueStateTokenMint ::
...
+    # pto (pfromJust #$ plookup # thisPolicy # pto mint)
```
SeungheonOh, line 66: Is it okay for this function to throw `perror`?

chfanghr: It may look ugly but failing is the intended behavior.

SeungheonOh: Hmm, okay. Since it's returning bool, I thought it shouldn't be partial.

src/Plutarch/Extra/StateThread.hs
peter-mlabs: Good call, done.

PR #142, opened 2022-12-10
JSON fixes, actually

cjay:
My previous versions were very wrong, sorry.
Tested in branch `cjay/json-fixes-working` (without liqwid-nix 2). Not tested in this branch here, because my nix doesn't seem to terminate (without doing any significant CPU load).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 This format is based on [Keep A Changelog](https://keepachangelog.com/en/1.0.0).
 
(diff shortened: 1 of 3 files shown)


peter-mlabs:
CI was busted on this due to redundant imports. These changes also include LN2.0, which means we'd need LN2.0 on `liqwid-onchain` to incorporate. I'm trying to figure out what the status is on LN2.0, but we may need to go back to an earlier version of this repo as a work around.

